PR #10, opened 2023-02-24
Automatic Release

sevenautumns:
This pull request introduces automatic Release creation on new tags.


wucke13:
Thank you! A few comments: 

1.I think we should only release if the branch is `main` __and__ when the tag starts with `v`. 
2. We should not ship the `rp` script externally for the static docker container

wucke13 (approved):

PR #23, opened 2023-03-03
man: add rosenpass.1

emilengler:
This commit adds a manual page for the rosenpass(1) utility. Currently, it's content is mostly a replica of rosenpass's built-in help.

### DIFF
--- a/doc/rosenpass.1
+++ b/doc/rosenpass.1
@@ -1,0 +1,105 @@
+.Dd $Mdocdate$
+.Dt ROSENPASS 1
+.Os
+.Sh NAME
+.Nm rosenpass
+.Nd builds post-quantum-secure VPNs
+.Sh SYNOPSIS
+.Nm
+.Op COMMAND
+.Op Ar OPTIONS ...
+.Op Ar ARGS ...
+.Sh DESCRIPTION
+.Nm
+performs cryptographic key exchanges that are secure against quantum-computers
+and outputs the keys.
+These keys can then be passed to various services such as wireguard or other
+vpn services as pre-shared-keys to achieve security against attackers with
+quantum computers.
+.Pp
+This is a research project and quantum computers are not thought to become
+practical in less than ten years.
+If you are not specifically tasked with developing post-quantum secure systems,
+you probably do not need this tool.
+.Ss COMMANDS
+.Bl -tag -width Ds
+.It Ar keygen private-key <file-path> public-key <file-path>
+Generate a keypair to use in the exchange command later.
+Send the public-key file to your communication partner and keep the private-key
+file secret!
+.It Ar exchange private-key <file-path> public-key <file-path> [ OPTIONS ] PEERS
+Start a process to exchange keys with the specified peers.
+You should specify at least one peer.
+.Pp
+It's
+.Ar OPTIONS
+are as follows:
+.Bl -tag -width Ds
+.It Ar listen <ip>[:<port>]
+Instructs
+.Nm
+to listen on the specified interface and port.
+By default
+.Nm
+will listen on all interfaces and select a random port.
+.It Ar verbose
+Extra logging.
+.El
+.El
+.Ss PEER
+Each
+.Ar PEER
+is defined as follows:
+.Qq peer public-key <file-path> [endpoint <ip>[:<port>]] [preshared-key <file-path>] [outfile <file-path>] [wireguard <dev> <peer> <extra_params>]
+.Pp
+Providing a
+.Ar PEER
+instructs
+.Nm
+to exchange keys with the given peer and write the resulting PSK into the given
+output file.
+You must either specify the outfile or wireguard output option.
+.Pp
+The parameters of
+.Ar PEER
+are as follows:
+.Bl -tag -width Ds
+.It Ar endpoint <ip>[:<port>]
+Specifies the address where the peer can be reached.
+This will be automatically updated after the first successful key exchange with
+the peer.
+If this is unspecified, the peer must initiate the connection.
+.It Ar preshared-key <file-path>
+You may specify a pre-shared key which will be mixed into the final secret.
+.It Ar outfile <file-path>
+You may specify a file to write the exchanged keys to.
+If this option is specified,
+.Nm
+will write a notification to standard out every time the key is updated.
+.It Ar wireguard <dev> <peer> <extra_params>
+This allows you to directly specify a wireguard peer to deploy the
+pre-shared-key to.
+You may specify extra parameters you would pass to
+.Qq wg set
+besides the preshared-key parameter which is used by
+.Nm .
+This makes it possible to add peers entirely from
+.Nm .
+.El
+.Sh EXIT STATUS
+.Ex -std
+.Sh SEE ALSO
+.Xr rp 1 ,
+.Xr wg 1
+.Sh STANDARDS
+This tool is the reference implementation of the Rosenpass protocol, written
+by Karolin Varner, Benjamin Lipp, Wanja Zaeske, and Lisa Schmidt.
+.Sh AUTHORS
+Rosenpass was created by Karolin Varner, Benjamin Lipp, Wanja Zaeske,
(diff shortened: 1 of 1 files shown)


emilengler:
I see, there's a file called `src/usage.md`. Maybe this one should be used to generate the manual page? :thinking:

emilengler:
cc @wucke13

koraa:
This is fantastic! Thank you very much!

I do think generating this from the usage.md file would be the best way to go; we should probably reformat usage.md to be proper markdown too…

emilengler:
Okay, let me see what I can do! :-)

emilengler:
How about generating `usage.md` from the manual page instead?
Converting manual pages to a different format is easy.
However, converting a different format to `mdoc(7)` often generates semantically bad man-pages...

koraa:
We could just call `man rosenpass` from within rosenpass, that would work.

Although I find the mdoc format quite unreadable. Can we clean that up somehow?

emilengler:
> We could just call man rosenpass from within rosenpass, that would work.

👍 

> Although I find the mdoc format quite unreadable. Can we clean that up somehow?

`mdoc(7)` is just like `man(7)` a macro language for the very old troff typesetting system invented at AT&T in the 70s. It definetly looks very obscure to people who are not used to writing manuals. Unfortunately, I am not aware of an easier way. The GNU project once tried to replace it with LaTeX (see GNU info) but they failed with it.

An alternative might be to use a tool, that is able to convert Markdown to either `mdoc(7)` or `man(7)`. One such tool for example is [lowdown](https://kristaps.bsd.lv/lowdown). It's author does not recommend this however.

man/rosenpass.1
```diff
@@ -0,0 +1,115 @@
...
+.Ex -std
+.Sh SEE ALSO
+.Xr rp 1 ,
+.Xr wg 1
+.Sh STANDARDS
+This tool is the reference implementation of the Rosenpass whitepaper, written
+by Karolin Varner, Benjamin Lipp, Wanja Zaeske, and Lisa Schmidt.
+.Sh AUTHORS
+Rosenpass was created by Karolin Varner, Benjamin Lipp, Wanja Zaeske,
+Marei Peischl,Stephan Ajuvo, and Lisa Schmidt.
```
wucke13, line 110: maybe put a space after the comma? And I believe Ajuvo is not the surname, @ajuvo how do you want to be mentioned here?

man/rosenpass.1
```diff
@@ -0,0 +1,115 @@
...
+.El
+.El
+.Sh EXIT STATUS
+.Ex -std
+.Sh SEE ALSO
+.Xr rp 1 ,
+.Xr wg 1
+.Sh STANDARDS
+This tool is the reference implementation of the Rosenpass whitepaper, written
+by Karolin Varner, Benjamin Lipp, Wanja Zaeske, and Lisa Schmidt.
```
wucke13, line 107: I believe the grammar here does not make clear whether you are crediting the authors of the whitepaper or of the tool. Maybe two sentences?

emilengler: See the AUTHORS section.

wucke13: I guess that settles it :smile:

doc/rosenpass.1
```diff
@@ -0,0 +1,106 @@
+.Dd $Mdocdate$
+.Dt ROSENPASS 1
+.Os
+.Sh NAME
+.Nm rosenpass
+.Nd builds post-quantum-secure VPNs with WireGuard
```
koraa, line 6: Can we mention that it performs a key exchange specifically and that it can work without WG?

doc/rosenpass.1
```diff
@@ -0,0 +1,106 @@
...
+These keys can then be passed to various services such as wireguard or other
+vpn services as pre-shared-keys to achieve security against attackers with
+quantum computers.
+.Pp
+This is a research project and quantum computers are not thought to become
+practical in less than ten years.
+If you are not specifically tasked with developing post-quantum secure systems,
+you probably do not need this tool.
+.Pp
+Each
```
koraa, line 25: Unfinished sentence

doc/rosenpass.1
```diff
@@ -0,0 +1,106 @@
...
+.Ss COMMANDS
+.Bl -tag -width Ds
+.It Ar keygen private-key <file-path> public-key <file-path>
+Generate a keypair to use in the exchange command later.
+Send the public-key file to your communication partner and keep the private-key
+file a secret!
```
koraa, line 31: a secret -> secret

doc/rosenpass.1
```diff
@@ -0,0 +1,106 @@
...
+This makes it possible to add peers entirely from
+.Nm .
+.El
+.Sh EXIT STATUS
+.Ex -std
+.Sh SEE ALSO
+.Xr rp 1 ,
+.Xr wg 1
+.Sh STANDARDS
+This tool is the reference implementation of the Rosenpass whitepaper, written
```
koraa, line 97: whitepaper -> protocol

doc/rosenpass.1
```diff
@@ -0,0 +1,106 @@
...
+by Karolin Varner, Benjamin Lipp, Wanja Zaeske, and Lisa Schmidt.
+.Sh AUTHORS
+Rosenpass was created by Karolin Varner, Benjamin Lipp, Wanja Zaeske,
+Marei Peischl, Stephan Ajuvo, and Lisa Schmidt.
+.Pp
+This manual page was written by
+.An Emil Engler
+.Sh BUGS
+Yes.
```
koraa, line 106: Please link the issue tracker

koraa (changes requested):
I think this is for the most part good enough. Thank you very much for your pull request, great work!

Thank you for your explanation, I still do not like mdoc, but given your explanation I think using it is out best option.

Before we can merge this, I think we should:

* [ ] Update the rosenpass command to display the manpage instead of usage.md
* [ ] Provide rp(1)
* [ ] Integrate the mangpages into the install process (CI, and rust)
* [ ] (Optionally) provide `rg help` that displays the rg man page

Thank you for your contribution; I am really glad you picked up creating the manpages, this is definetly an area our current code falls short in.

doc/rosenpass.1
emilengler: How would you phrase it? I can only come up with formulations, that would make rosenpass sound like it's own VPN protocol.

emilengler:
>  Update the rosenpass command to display the manpage instead of usage.md

Would you launch the `man(1)` program or do something like `curl --manual`?

emilengler:
The suggestions you made would include a few modifications to the cargo setup. Most notably:

- Let `cargo install` respect the manual page
- Potentially compile the manual page into ASCII and then include it statically, the same way we do with the Markdown file

@wucke13: Any thoughts on this? I know how to solve this with a Makefile, but not with cargo. Is it possible without a dirty hack?

doc/rosenpass.1
wucke13: > enhances WireGuard with post-quantum secure key-exchange

?

wucke13:
I'd take a look at how https://github.com/BurntSushi/ripgrep does it, basically they put the manpage to the `docs/` folder and they added a text-to-text conversion in the `build.rs` to fix this. Here is another manpage example: https://github.com/ClementTsang/bottom/blob/master/build.rs . So something akin this would be good. I'm not sure how to make it work with `cargo install` though.

koraa:
> Would you launch the `man(1)` program or do something like `curl --manual`?

I'd launch man(1). Much easier :)

koraa:
I've been fairly unresponsive these past couple of weeks since I was traveling. Thank you for still sticking by this work, it is much appreciated :pray:

emilengler:
> 



> Merging for now. Thank you for your contribution! Could you please take the time to address the open issues in a separate PR?

I am currently a bit busy. I will try to do so in the next week or so.

koraa:
Thank you!

PR #37, opened 2023-04-04
rename protocol::Server -> protocol::CryptoServer

wucke13:
I think it makes sense to be more specific in the type name. The distinction between CryptoServer and AppServer is quite nice, so lets formalize it :)


wucke13:
closing in favor of #38

PR #53, opened 2023-04-13
invoke `man(1)` when requesting help

emilengler:
This commit invokes the `man(1)` command with the rosenpass manual page, when invoking `help`.

In the future, we should handle the case, that a rosenpass manual page is not installed, somewhat better. My idea would be to render the manual page to an HTML version regularly and provide a link to it inside the error message.

PR #54, opened 2023-04-13
doc: add a manual page for `rp(1)`

emilengler:


koraa (approved):

koraa:
\o/ Thank you! You're going to add a build system integration right`

emilengler:
> You're going to add a build system integration right?

What do you exactly mean by this? A simple invocation of a man-page renderer to see if the file is valid? A way to install it?

koraa:
> What do you exactly mean by this? A simple invocation of a man-page renderer to see if the file is valid? A way to install it?

Hi @emilengler, again thank you for writing all the man pages.

Yes, a way to install it!

Upon `cargo install rosenpass`, the man pages should be available system wide. I.e. it needs to be in the right place for `man rosenpass` and `man rp` to work. This is especially important in connection with #53 since this invokes `man rosenpass`.

> A simple invocation of a man-page renderer to see if the file is valid?

I think that would be CI integration which would be wonderful!

PR #56, opened 2023-04-15
major rewrite of application server & frontend

### LINKED ISSUE #27: Support for multiple listen sockets
I tried following this setup: https://rosenpass.eu/#start

Server: 192.168.2.107/24
<pre>
rp exchange 192.168.2.107-sec dev rosenpass0 listen 192.168.2.107:9999 peer 192.168.2.106-pub allowed-ips 192.168.123.0/24
</pre>

Client: 192.168.2.106/24
<pre>
rp exchange 192.168.2.106-sec dev rosenpass0 peer 192.168.2.107-pub endpoint 192.168.2.107:9999 allowed-ips 192.168.123.0/24
</pre>

However, I am getting this output from the client:
<pre>
[2023-03-14T20:59:57Z ERROR rosenpass] unexpected error after processing 0 messages: Address family not supported by protocol family (os error 47) disabled backtrace
[2023-03-14T20:59:57Z ERROR rosenpass] reinitializing networking in 0.01! 10 tries left.
[2023-03-14T20:59:57Z ERROR rosenpass] unexpected error after processing 0 messages: Address family not supported by protocol family (os error 47) disabled backtrace
[2023-03-14T20:59:57Z ERROR rosenpass] reinitializing networking in 0.02! 9 tries left.
[2023-03-14T20:59:58Z ERROR rosenpass] unexpected error after processing 0 messages: Address family not supported by protocol family (os error 47) disabled backtrace
[2023-03-14T20:59:58Z ERROR rosenpass] reinitializing networking in 0.04! 8 tries left.
[2023-03-14T20:59:59Z ERROR rosenpass] unexpected error after processing 0 messages: Address family not supported by protocol family (os error 47) disabled backtrace
[2023-03-14T20:59:59Z ERROR rosenpass] reinitializing networking in 0.08! 7 tries left.
...
</pre>

To my understanding this is because the socket the rosenpass binary is using for sendig is somehow bound for IPv6:
<pre>
 98150 rosenpass CALL  socket(AF_INET6,0x8002<SOCK_DGRAM|SOCK_CLOEXEC>,0)
 98150 rosenpass RET   socket 3
 98150 rosenpass CALL  bind(3,0x7f7ffffd19a8,28)
 98150 rosenpass STRU  struct sockaddr { AF_INET6, [::]:0 }
 98150 rosenpass RET   bind 0
...
 98150 rosenpass CALL  sendto(3,0x7f7ffffd24b0,0x444,0x400<MSG_NOSIGNAL>,0x7f7ffffd1a30,0x10)
 98150 rosenpass STRU  struct sockaddr { AF_INET, 192.168.2.107:9999 }
 98150 rosenpass RET   sendto -1 errno 47 Address family not supported by protocol family
 98150 rosenpass CALL  write(2,0x878b524c500,0xd9)
 98150 rosenpass GIO   fd 2 wrote 217 bytes
       "\^[[0m\^[[38;5;8m[\^[[0m2023-03-14T20:59:57Z \^[[0m\^[[1m\^[[31mERROR\
        \^[[0m rosenpass\^[[0m\^[[38;5;8m]\^[[0m unexpected error after proces\
        sing 0 messages: Address family not supported by protocol family (os e\
        rror 47) disabled backtrace
       "
 98150 rosenpass RET   write 217/0xd9
 98150 rosenpass CALL  write(2,0x878b524c500,0x8c)
 98150 rosenpass GIO   fd 2 wrote 140 bytes
       "\^[[0m\^[[38;5;8m[\^[[0m2023-03-14T20:59:57Z \^[[0m\^[[1m\^[[31mERROR\
        \^[[0m rosenpass\^[[0m\^[[38;5;8m]\^[[0m reinitializing networking in \
        0.01! 10 tries left.
       "
</pre>

<strong>
This unwrap_or makes the socket IPv6 since `listen` is not set for the client.
</strong>

To fix this, the client should bind a v4 socket too and depending on endpoint it should use that socket for sending.
Alternatively it could just try the v4 socket in case the v6 one failed.

I will look into implementing that but I have little knowledge on rust.

https://github.com/rosenpass/rosenpass/blob/main/src/main.rs#L356
Adding `listen` for the client makes it work with IPv4.

---

Because the client is trying to send via an IPv6 socket, I tried just using IPv6 to connect to the server:

Server:
<pre>
rp exchange 192.168.2.107-sec dev rosenpass0 listen fe80::fce1:bbff:fed1:3a06%vio0:9999 peer 192.168.2.106-pub allowed-ips 192.168.123.0/24
</pre>

Client:
<pre>
rp exchange 192.168.2.106-sec dev rosenpass0 peer 192.168.2.107-pub endpoint fe80::fce1:bbff:fed1:3a06%vio0:9999 allowed-ips 192.168.123.0/24
</pre>

Now everything works, I can assign IP addresses to the wg/rosenpass interfaces and ping through them.

I had to adjust the rp script sligthly to make it work with OpenBSD, but I did not add any changes to the rosenpass binary:
https://github.com/moritzbuhl/mystuff/tree/master/net/rosenpass

wucke13:
- adds TOML based configuation files
  - with example configuratios in config-examples
- reimplments arcane CLI argument parser as automaton
- adds a new CLI focused arround configuration files
- moves all file utility stuff from `main.rs` to `util.rs`
- moves all AppServer stuff to dedicated `app_server.rs`
- add mio for multi-listen-socket support (should fix #27)
- consistency: rename private to secret

@koraa sorry.


wucke13:
This is a first draft for how I would like the config file to be, and a bunch of other things that I wanted to improve anyways. This blows the dependency tree quite a lot, however further down the roadmap I'd like to see the all cli/config complexity to be removed from `rosenpass` and moved into `rp` (which by then would be rewritten in Rust, as demanded by #1 ).

config-examples/peer-a-config.toml
```diff
@@ -0,0 +1,18 @@
+public_key = "peer-a-public-key"
```
koraa, line 1: Can we come up with a naming scheme that aligns with the whitepaper?

config-examples/peer-a-config.toml
```diff
@@ -0,0 +1,18 @@
...
+secret_key = "peer-a-secret-key"
+listen = ["[::]:10001"]
```
koraa, line 3: Do we already support multiple sockets?

config-examples/peer-b-config.toml
```diff
@@ -0,0 +1,18 @@
+public_key = "peer-b-public-key"
+secret_key = "peer-b-secret-key"
+listen = ["[::]:10002"]
+verbosity = "Quiet"
+
+[[peers]]
```
koraa, line 6: Can you add a second peer in an example?

config-examples/peer-b-config.toml
```diff
@@ -0,0 +1,18 @@
+public_key = "peer-b-public-key"
+secret_key = "peer-b-secret-key"
+listen = ["[::]:10002"]
+verbosity = "Quiet"
```
koraa, line 4: Is that case sensitive?

Cargo.toml
```diff
@@ -26,6 +25,10 @@ thiserror = "1.0.38"
 paste = "1.0.11"
 log = { version = "0.4.17", optional = true }
 env_logger = { version = "0.10.0", optional = true }
+serde = { version = "1.0.160", features = ["derive"] }
```
koraa, line 28: That is a lot of packets. We should make it a goal to reduce the external crate footprint again over time.

wucke13: Indeed, I plan to eventually move most of this to the frontend tool

config-examples/peer-b-config.toml
wucke13: I think so. I believe we should make this be the different log `levels`, `trace`, `debug`, `info`, `warn`, `error`, `true` -> `info`, `false` -> `warn`,`verbose`-> `debug`

config-examples/peer-b-config.toml
wucke13: I never tried _yet_ :smile:

config-examples/peer-a-config.toml
wucke13: Kind of. There is a couple of caveats. Currently, we are always responding from the first socket, which will error on ipv4/ipv6 problems, but that is already a TODO

config-examples/peer-a-config.toml
wucke13: you mean `spki`, `sski`, `spkr`, `sskr`? I'm mostly complient, just that I would like to remove the `i`/`r` suffix because you also asked for a more than two peers setup. So maybe `peer-[a..c]-ssk` and `peer-[a..c]-spk` and `peer-[a..c]-osk`?

koraa: We use `sskm` for `static-secret-key-mine` and `static-secret-key-theirs`. If we use `peer-.*` and `mine-.*` we would need to adjust the whitepaper for compatibility.

koraa: Though honestly, I am a fan of establishing the four character naming scheme and just using it very consistently. I know its not super intuitive on first use but we need keys in so many places…

config-examples/peer-a-config.toml
koraa: So you would consider this preparatory work for multiple udp sockets?   Would it be hard to just add that as part of this PR?

config-examples/peer-b-config.toml
koraa: Ok sounds good. Can we make it case insensitive?

koraa:
AFAICS this PR is mostly ready to merge lest for two blockers:

* [ ] Responding on the correct socket
* [ ] Naming scheme consistency with the sski/sskr naming scheme.

wucke13:
>  Responding on the correct socket

This turns out to be surprisingly tricky.

1. We don't know if we are on a dual-stack OS, and there is no way of finding out, which is generic & robust.
2. If we are responder, we can always just respond via the socket that received the message we are responding to. This requires a `HashMap<SocketAddr, usize>`, where the `SocketAddr` is IP & port of the Rosenpass on a peer's machine and `usize` is the index of the local socket in the `AppServers`'s socket vector.
3. If we are initiator, we don't know which socket to send from :angry:. Why is that?
4. If we are on a non-dual stack OS, the IP Address Family of the peer's `SocketAddr` and the local socket must be equal.
5. Because of 4., we can not just pick any socket.
6. However, if we are on a dual-stack OS, we can (and should!) use just one socket, even if its Address Family is IPv4 it can send to IPv6 peers.

koraa:
>     2. If we are responder, we can always just respond via the socket that received the message we are responding to. This requires a `HashMap<SocketAddr, usize>`, where the `SocketAddr` is IP & port of the Rosenpass on a peer's machine and `usize` is the index of the local socket in the `AppServers`'s socket vector.

If that's what we got to do, that's what we got to do. I think we will end up with an address type that is a tuple to some socket handle and the ip address used within the socket. I wonder can't try_recv just return the socket and address it received on in such an address?

>     3. If we are initiator, we don't know which socket to send from angry. Why is that?

We have to try them all…round-robin unless a specific socket is found that generates a successful handshake.

I think that nets us a new address Enum with N different possible state:

1. Socket unknown; just stores the string address and a round-robin handle to the Socket to use (using interior mutability)
2. Socket known; one arm for each possible socket/address type and a handle indicating the specific socket.

wucke13:
> We have to try them all…round-robin

This is implemented. The rest we have to do eventually, but let's first get this merged.

koraa:
@wucke13 Please give my updates another review! I'll give this another look myself some time later and then I think we can merge.

config-examples/peer-a-config.toml
wucke13: In general I agree. However, `theirs` and `mine` only works well if there is exactly two peers in total, and its a relative naming which would not work well when both peers are executed on the same host, as for example a mini-demo.

koraa (approved):

PR #75, opened 2023-05-23
fix: Show cargo fmt errors

koraa:


koraa:
@wucke13 Please tell me async if this is the right adjustment for the nix file. I am merging now because I want the main branch green :)

flake.nix
```diff
@@ -325,7 +325,7 @@
           checks = {
             cargo-fmt = pkgs.runCommand "check-cargo-fmt"
               { inherit (self.devShells.${system}.default) nativeBuildInputs buildInputs; } ''
-              cargo fmt --manifest-path=${./.}/Cargo.toml --check > $out
+              cargo fmt --manifest-path=${./.}/Cargo.toml --check && touch $out
```
emilengler, line 328: I think you might be looking for this: https://stackoverflow.com/questions/418896/how-to-redirect-output-to-a-file-and-stdout

koraa: I don't think we need the output in the file…what do you say @wucke13

PR #78, opened 2023-05-23
Fix keys going out of sync due to a race condition

koraa:


rp
```diff
@@ -44,9 +44,14 @@ dbg() {
 }
 
 
-add_warning_tag() {
+detect_git_dir() {
```
emilengler, line 47: I think you've based your commit on the wrong branch. This change looks a bit out of the place here.

src/protocol.rs
```diff
@@ -657,7 +659,22 @@ impl Mortal for SessionPtr {
     }
 
     fn retire_at(&self, srv: &CryptoServer) -> Option<Timing> {
-        self.created_at(srv).map(|t| t + REKEY_AFTER_TIME)
+        // If we where the initiator, wait an extra ten seconds to avoid
```
emilengler, line 662: typo: `s/where/were/g`

src/protocol.rs
```diff
@@ -657,7 +659,22 @@ impl Mortal for SessionPtr {
...
+        use HandshakeRole::*;
+        let role = self.get(srv).as_ref().map(|p| p.handshake_role);
+        let extra_wait = match role {
+            Some(Initiator) => 10.0,
+            Some(Responder) => 0.0,
+            None => 0.0,
+        };
+
+        self.created_at(srv)
+            .map(|t| t + REKEY_AFTER_TIME + extra_wait)
```
emilengler, line 677: If we "change" the value of `REKEY_AFTER_TIME`, wouldn't it make sense to make this constant private and change its name then? From a software engineering point of view, constants imply to me somewhat of an absolute truth, which wouldn't be true with this patch anymore.

I would suggest changing this constant's name to something like "MIN_REKEY_AFTER_TIME" and maybe making it private.

src/protocol.rs
```diff
@@ -1623,9 +1641,29 @@ impl CryptoServer {
             // ICR7
             peer.session()
                 .insert(self, core.enter_live(self, HandshakeRole::Responder)?)?;
+            // TODO: This should be part of the protocol specification.
+            peer.hs().take(self);
```
emilengler, line 1645: I do not understand how this is related to the race condition.

rp
koraa: Nope; I intentionally added both commits. Creating two PRs is a bunch of extra work and it does no harm :shrug:

src/protocol.rs
koraa: Thats a bit of a bike shed, but ok :shrug:

src/protocol.rs
koraa: Thank you!

src/protocol.rs
koraa: This is what actually fixes the race condition; it aborts any ongoing handshake from initiator perspective.

PR #90, opened 2023-06-09
CLI improvements and CI refinery

wucke13:

PR #97, opened 2023-06-30
fix devshell bug introduced in #90

emilengler:
This commit reverts the CMake change done in #90, as it broke the invocation of `cargo build`.

While this might be considered as a hotfix, it at least fixes building through `cargo build` and, in my opinion more importantly, makes it possible to run unit tests through `cargo test` again.


wucke13:
Don't merge this yet, instead cherry pick 5bebdd9284a56e65136803981adc3dd29769ed6c

wucke13 (approved):

PR #102, opened 2023-07-19
Readme: Add installation and compilation guide

marek22k:


marek22k:
> I've installed it via: `cargo install --git https://github.com/rosenpass/rosenpass`

Then where does it copy the executable file? Does it install rp with it?
If something like this was on the server, it might be a hindrance, as it should then be accessible to all users - i.e. in `/usr/local/bin`. (And I personally would not want to run a compiler with internet access as root).

koraa:
Sorry for the long delay in reviewing.

The PR sounds good to me; I don't think the guide belongs in the readme though. It is specific to ubuntu and how to install a rust crate is a standard procedure.

Could you please create a wiki directory instead and put the guide under wiki/installation/ubuntu/readme.md and create wiki/readme.md and wiki/installation/readme.md ?

marek22k:
It is rather Debian than Ubuntu specific (Ubuntu is based on Debian). I'm happy to separate the instructions, however I would link to it from the README.

marek22k:
What about the access rights? What about a mirror to notabug.org and GitLab? Both hosting platforms also have a wiki function. Is it possible to see the Git history from the wiki online (for example as a separate Git repo with PR and issue functions)?

Horever, I like the idea of separating code and documentation.

koraa:
> @koraa GitHub has a wiki function built in. I'm in favor of having this documentation in that wiki. I want to encourage frequent updates of the wiki, but I want to avoid a lot of noise in the commit history of the repo itself.
> 
> I do foresee that you would like to have the wiki available via git, and fortunately that is possible: the wiki on a GitHub project is its own repo, available like this: `git clone https://github.com/rosenpass/rosenpass.wiki.git`.

I am happy to use the GH wiki functionality…

koraa:
> > @koraa GitHub has a wiki function built in. I'm in favor of having this documentation in that wiki. I want to encourage frequent updates of the wiki, but I want to avoid a lot of noise in the commit history of the repo itself.
> > I do foresee that you would like to have the wiki available via git, and fortunately that is possible: the wiki on a GitHub project is its own repo, available like this: `git clone https://github.com/rosenpass/rosenpass.wiki.git`.
> 
> I am happy to use the GH wiki functionality…

Although, we should move to a platform-independent, git-backed wiki at some point (soon).

marek22k:
> Although, we should move to a platform-independent, git-backed wiki at some point (soon).

Well, we could just create an extra Git repo for the wiki. This would make it platform independent. The wiki tab could perhaps be redirected to the other repo. GitHub, notabug.org and GitLab have the option to display Markdown in the repo outside of a wiki.

emilengler:
We already heavily rely on GitHub. Using their wiki makes no difference here antmore. In order to fully participate in the development, one has to read content from GitHub. Even if that is just a reference to an issue/PR from within a commit message.

koraa:
@emilengler OK! Good point. Then let's do that.

koraa:
> @marek22k Again, thank you for the contribution. After the long pause, I finally implemented the proposal of [adding this to the Wiki](https://github.com/rosenpass/rosenpass/wiki/Compilation-&-Installation). With that, I'm closing this PR. You are most welcome to add anything missing to the Wiki!

Thank you for doing ths!

PR #104, opened 2023-07-19
rp: Defaults to dual-stack

marek22k:
If currently no IP address, only on IPv6 is listen by default. This commit would make it listen dual-stack - i.e. IPv4 and IPv6 - by default.


koraa (approved):

koraa:
Thank you for this pull request!

PR #127, opened 2023-09-14
Implement test coverage checks

emilengler:
This PR adds a script `maint/coverage.sh` that provides a wrapper around `grcov` in order to execute unit tests.

For now, it only outputs in lcov, although many other formats are available as well.

Below, there is an HTML output (maybe we can integrate it to our website):
Screenshot 2023-09-14 at 11 14 43

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,7 @@ _markdown_*
 **/result
 **/result-*
 .direnv
+
+# Test coverage
+/coverage
+/lcov.info
--- a/maint/coverage.sh
+++ b/maint/coverage.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+#
+# In order use this, you need to have `grcov` in your path and `llvm-tools` in
+# your Rust toolchain.
+#
+# ```sh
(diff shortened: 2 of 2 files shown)


emilengler:
@stv0g I've briefly looked at your suggestion. While it is probably a very nice tool, I've personally got the impression that it is more meant to be used in a SAAS fashion. I personally prefer to be able to run the test coverage locally.

wucke13:
@emilengler I feel guilty for wasting your time by not documenting well, what we already have. You can literally run `nix develop .#coverage cargo llvm-cov` and you are already there. Without Nix one of course has to get `cargo-llvm-cov` from somewhere...

What however would be awesome is better integration with CI, e.g. running the coverage check in CI, and reporting the result somehow to the respective PR.

Please have a look if `cargo-llvm-cov` is inferior to your script in any way, if so I'm happy to merge. If not, I would prefer using the external, already existing tool, instead of adding more scripts to this repo.

emilengler:
`llvm-cov` seems fine to me.

Issue #137, opened 2023-10-03
Implement the cookie mechanism


koraa:
Implement the cookie mechanism as described in the [wireguard whitepaper](https://www.wireguard.com/papers/wireguard.pdf). Document this appropriately in the whitepaper.

TouchstoneTheGreat:
@koraa Idk this desc. is enough or not. But I would like get more details on this !
> Implement the cookie mechanism as described in the [wireguard whitepaper](https://www.wireguard.com/papers/wireguard.pdf). Document this appropriately in the whitepaper.

koraa:
> @koraa Idk this desc. is enough or not. But I would like get more details on this !
> 
> > Implement the cookie mechanism as described in the [wireguard whitepaper](https://www.wireguard.com/papers/wireguard.pdf). Document this appropriately in the whitepaper.

Sure! Could you please read sections 5.3, 5.4.4, 5.4.7, and 6.6 of the wireguard whitepaper? The cookie reply mechanism is described in detail there. The same mechanism should be implemented in Rosenpass and the mechanism should be described in our whitepaper too.

TouchstoneTheGreat:
If you are maintainer I want ask is this good approach

1. Generate a random cookie for each handshake.
2. Send the cookie to the initiator in the first handshake message.
3.Compute a MAC of the cookie using the initiator's private key and send the MAC back to the responder in the second 4.handshake message.
5.Verify the MAC using the initiator's public key in the responder.

Can you suggest any other approaches ?

koraa:
> If you are maintainer I want ask is this good approach
> 
>     1. Generate a random cookie for each handshake.
> 
>     2. Send the cookie to the initiator in the first handshake message.
>        3.Compute a MAC of the cookie using the initiator's private key and send the MAC back to the responder in the second 4.handshake message.
>        5.Verify the MAC using the initiator's public key in the responder.
> 
> 
> Can you suggest any other approaches ?

Did you take that from the whitepaper? I think there are some gaps in the plan if this sketch was supposed to be comprehensive. I would start by coming up with a rule deciding under which conditions the cookie mechanism will activate. It should not be active by default, only under contention.

prabhpreet:
/attempt #137

[...]

PR #142, opened 2023-10-24
Add cookie mechanism

prabhpreet:
Issue #137 @koraa 

Currently this PR has edits to the whitepaper for implementation to add cookie mechanism similar to the Wireguard whitepaper. Major differences are using the tree-based domain separated `hash` functions used by the `mac` and `cookie` fields in the envelope.

To trigger a cookie reply, the Wireguard whitepaper does not hint at specific criteria to determine an under load condition. I took a look at the linux kernel module wireguard implementation, and userspace implementation in rust. This psuedocode represents how the implementations determine `under_load` condition to trigger a cookie reply:

Initialization: 
```
last_under_load: Option<Time> = None;
under_load: bool = false;
```

On receiving a packet:
```
under_load =  handshake_queue_len >= MAX_QUEUED_INCOMING_HANDSHAKES_THRESHOLD;

if(under_load) {
    last_under_load = Some(time::now());
}
else if let Some(last_under_load) =  last_under_load {
    under_load = (time::now() < last_under_load + Duration::from_secs(LAST_UNDER_LOAD_WINDOW));
    if(!under_load) {
        last_under_load = None;
    }
}
```


koraa:
@prabhpreet Thank you for submitting this. Your updates in the whitepaper seem right on track!

koraa:
@prabhpreet Thank you! I did not perform in in-depth review yet but a very quick glace made me  think that this is great code!

Would you mind adding a section to the whitepape about how rosenpass now decides it is under load?

prabhpreet:
I surely can- however, wirepass's doesn't specify or bound an implementation of the under load condition, and leaves that for the client implementations to gauge the under load condition for themselves. 

I can also add comments in the source code instead on how the under_load condition works for `AppServer`. Then, it could be mentioned in the whitepaper that it is up to the client to gauge the under load condition for themselves to trigger the cookie DoS mechanism.

Let me know what is preferable.

koraa:
> @koraa I surely can- however, wirepass's doesn't specify or bound an implementation of the under load condition in its whitepaper, and leaves that for the client implementations to gauge the under load condition for themselves.
> 
> I can also add comments in the source code instead on how the under_load condition works for `AppServer`. Then, it could be mentioned in the whitepaper that it is up to the client to gauge the under load condition for themselves to trigger the cookie DoS mechanism.
> 
> Please let me know what is preferable.

I think we should add the info to the whitepaper as an informative section but also communicate that its up to the implementing application to choose when to use the cookie mechanism…

Thank you for your dedication to working on this task; this is really coming together nicely :)

prabhpreet:
Thanks! Sounds good, will surely add changes to the whitepaper.

prabhpreet:
@koraa Need your/team's input here

I was revisiting the re transmission logic of Wireguard's whitepaper. Essentially, it states that the cookie mechanism does not trigger any retransmissions- rather, retransmissions happen as per its logic (Section 6), with the re-transmissions having a cookie appended to them. 

I have updated the whitepaper details accordingly, and added two tests- one when responder is under load, and one where initiator is under load.

However, the test for when the initiator is under load is facing an issue.

The test is as follows:

1. Initiator ---[InitHello] ---> Responder
2. Responder -----[RespHello] ----> Initiator under load
3. Initiator Under Load ----[Cookie Reply] ----> Responder
4. Initiator Under Load ----[Retransmits InitHello as per retransmission logic] -----> Responder
5. Responder ---- [RespHello with cookie] ----> Initiator under load
6. Continued with cookie value

Step 3 is where I'm facing an issue. The cookie reply message consists of a `sid` field which is used to locate the `Peer`. However during processing the `RespHello` state, the `Session` is not stored in `Peer` (rather stored after `InitConf` is processed.

This means that `sid` cannot be used to get `Peer` to store the cookie value which would be appended to subsequent messages. 

Here are some options I could think of for this solve:

1. Store `Session` of responder earlier in the exchange when receiving RespHello. Was this not stored intentionally due to the State Disruption Attack mechanism?
2. Use a different identifier unique to Peer (`PeerID` or a randomly generated value). Use this value instead of `sid` in the CookieReply message

Please let me know if there is a preferable route or some other way to solve this. Thanks

prabhpreet:
@koraa any thoughts?

koraa:
@prabhpreet Are you doing time tracking by the way?

koraa:
@prabhpreet The issue only arises with the initiator under load, right? Would just implementing the mechanism for responders under load be an option?

prabhpreet:
@koraa I'm not time tracking the hours on this.

That's correct - the problem only arises when the initiator is under load. 

I've already implemented the logic where the responder is under load. I can modify the test when the initiator is under load to ignore processing any messages, and bail with an error for traceability. The implication is that there would be no DoS mitigation for the initiator.

If that's sufficient for the time being, I can wrap up this PR for review with changes to the whitepaper.

koraa:
> @koraa I'm not time tracking the hours on this.
> 
> That's correct - the problem only arises when the initiator is under load.
> 
> I've already implemented the logic where the responder is under load. I can modify the test when the initiator is under load to ignore processing any messages, and bail with an error for traceability. The implication is that there would be no DoS mitigation for the initiator.
> 
> If that's sufficient for the time being, I can wrap up this PR for review with changes to the whitepaper.

Does WireGuard provide Initiator DOS mitigation in this way? I don't think it does.

The point of this cookie mechanism is to avoid IP spoofing attacks, we could just insert another `auth` token at the start of the RespHello packet…the expensive operations in RespHello processing are the decaps_and_mix operations, if the responder proves

koraa:
Sorry, misclicked.

prabhpreet:
> > @koraa I'm not time tracking the hours on this.
> > That's correct - the problem only arises when the initiator is under load.
> > I've already implemented the logic where the responder is under load. I can modify the test when the initiator is under load to ignore processing any messages, and bail with an error for traceability. The implication is that there would be no DoS mitigation for the initiator.
> > If that's sufficient for the time being, I can wrap up this PR for review with changes to the whitepaper.
> 
> Does WireGuard provide Initiator DOS mitigation in this way? I don't think it does.
> 

To my surprise, I think it does- the whitepaper seems to suggest so and so does the `wg_receive_handshake_packet` function in its Linux implementation:

image

https://git.zx2c4.com/wireguard-linux/tree/drivers/net/wireguard/receive.c#n92

> The point of this cookie mechanism is to avoid IP spoofing attacks, we could just insert another `auth` token at the start of the RespHello packet…the expensive operations in RespHello processing are the decaps_and_mix operations, if the responder proves they are in possession of a valid handshake state, there is very strong authentication through the static kem…
> 

I'd be curious into looking into this- I see you have created another issue. I may post some questions in the other task once I have clarity.

> I think implementing the cookie mechanism for initiators under load is good enough for now, can you finish the PR for just that case?

I've updated the code and the whitepaper for responders under load- ready for review!

koraa:
@prabhpreet Just a sidenote: Merging main into your branch is a workflow that makes it super hard to figure out what is going on…

I created #187 just to be able to review a clean state and I had to merge the current main branch into your branch, issue a `git reset origin/main`, use `git commit -aC cb97f9058150a748a14a4b811ca08bbb1a95eb22` to recover one of your commits and then I had to update the commit message to something fitting our `feat: …` template using `git commit --amend`.

Luckily, the merge was clean, so it just worked…

Anyway, you can reset your code to that branch for a cleaner slate.

koraa:
OK, looking over the code and the whitepaper, first of all: Great job!

papers/whitepaper.md
```diff
@@ -428,12 +428,89 @@ The responder code handling InitConf needs to deal with the biscuits and package
...
+mac = lhash("mac", spkt, MAC_WIRE_DATA)[0..16]
+```
+
+where `MAC_WIRE_DATA` represents all bytes of msg prior to `mac` field in the envelope.
+
+If a client receives an invalid `mac` value for any message, it will discard the message.
+
+#### Envelope cookie field
+
+The `cookie_tau` value encrypted as part of `cookie` field in the cookie reply message is decrypted by its receiver and stored as the `last_recvd_cookie` for a limited time (120 seconds). This value is then used by the sender to append a `cookie` field to the sender's message envelope to retransmit the handshake message.  This is the equivalent of Wireguard's `mac.2` field and is determined as follows:
```
koraa, line 466: >The `cookie_tau` value encrypted as part of `cookie` field in the cookie reply message is decrypted by its receiver and stored as the `last_recvd_cookie` for a limited time (120 seconds).

Again, can we be concrete here and refer to the initiator?

koraa, line 466: > This value is then used by the sender to append a `cookie` field to the sender's message envelope to retransmit the handshake message.

Which is confusing because here you are talking about the RespHello sender whereas before you where talking about the CookyReply receiver. Lets again just say initiator.

koraa, line 466: >  This value is then used by the sender to append a cookie field to the sender's message envelope to retransmit the handshake message.

Appending is not really what happens here – right?  The cookie field is always there, usually it just is set to some default value – zero I think?

papers/whitepaper.md
```diff
@@ -428,12 +428,89 @@ The responder code handling InitConf needs to deal with the biscuits and package
...
+
+```
+
+if (is_zero_length_bitstring(last_recvd_cookie)  ||  last_cookie_time_ellapsed >= 120) {
```
koraa, line 470: `is_zero_length_bitstring` is quite weird, nothing in rosenpass uses bit level stings…how about we just use an Option value?

papers/whitepaper.md
```diff
@@ -428,12 +428,89 @@ The responder code handling InitConf needs to deal with the biscuits and package
...
+    cookie = 0^16; //zeroed out 16 bytes bitstring
```
koraa, line 471: Lets just use rust notation. I think `cookie.zeroize()` would be quite clear…

papers/whitepaper.md
koraa, line 470: Why not call the cookie field `peer.cookie`?

papers/whitepaper.md
```diff
@@ -428,12 +428,89 @@ The responder code handling InitConf needs to deal with the biscuits and package
...
+```
+
+Here, `last_recvd_cookie` is the last received decrypted data of the `cookie` field from a cookie reply message by a hanshake message sender, `last_cookie_time_ellapsed` is the amount of time in seconds ellapsed since last cookie was received, and `COOKIE_WIRE_DATA` are the message contents of all bytes of the retransmitted message prior to the `cookie` field.
+
+The sender can use an invalid value for the `cookie` value, when the receiver is not under load, and the receiver must ignore this value.
+However, when the receiver is under load, it may reject messages with the invalid `cookie` value, and issue a cookie reply message. 
+
+### Conditions to trigger DoS Mechanism
+
+Rosenpass implementations are expected to detect conditions in which they are under computational load to trigger the cookie based DoS mitigation mechanism by replying with a cookie reply message.
```
koraa, line 485: Lets simplify this: `This whitepaper does not mandate any specific mechanism to detect responder contention that would trigger use of the cookie mechanism` or something similar…

koraa (changes requested):
Thank you for updating the whitepaper; I think it really helped me to gain a precise understanding of how you implemented the cookie mechanism.

The information you added is nice and exhaustive, I think we can simplify it a bit to make it easier for newcomers to digest the information.

Anyway, great job!

papers/whitepaper.md
```diff
@@ -428,12 +428,89 @@ The responder code handling InitConf needs to deal with the biscuits and package
...
+#### Cookie Reply Message
+
+The cookie reply message consists of the `sid` of the client under load, a random 24-byte bitstring `nonce` and encrypting `cookie_tau` into a `cookie` reply field which consists of the following (from the perspective of the cookie reply sender):
+
+```
+cookie_tau = lhash("cookie-tau",r_m, a_m)[0..16]
+cookie = XAEAD(lhash("cookie-key", spkm), nonce, cookie_tau , mac_peer)
+```
+
+where `r_m` is a secret variable that changes every two minutes to a random value. `a_m` is a concatenation of the source IP address and UDP source port of the client's peer. `cookie_tau` will result in a truncated 16 byte value from the above hash operation. `mac_peer` is the `mac` field of the peer's handshake message to which message is the reply.
```
koraa, line 450: Can you also specify how IP address and port are encoded for generating the cookie key?

koraa (changes requested):

rosenpass/src/app_server.rs
```diff
@@ -549,7 +563,37 @@ impl AppServer {
                 }
 
                 ReceivedMessage(len, endpoint) => {
-                    match self.crypt.handle_msg(&rx[..len], &mut *tx) {
+                    let msg_result = match self.under_load {
```
koraa, line 566: Could you factor that out into a new function?

rosenpass/src/app_server.rs
```diff
@@ -549,7 +563,37 @@ impl AppServer {
...
+                        DoSOperation::UnderLoad { last_under_load: _ } => {
+                            //TODO: Lookup peer through addresses (hash)
+                            let index = self
+                                .peers
+                                .iter()
+                                .enumerate()
+                                .find(|(_num, p)| {
+                                    if let Some(ep) = p.endpoint() {
+                                        ep.addresses() == endpoint.addresses()
```
koraa, line 575: Receiving a message from an unknown endpoint is actually all right…people would like to be able to connect to a VPN even when switching from mobile hotspot to office wifi :)

Also this triggers an expensive `O(n)` operation to find the address, which might actually be used for a DOS.

Lets just use the address in the IP packet…

The entire cookie mechanism is there to perform proof-of-ip-ownership, i.e. handle IP spoofing attacks.

rosenpass/src/app_server.rs
```diff
@@ -704,11 +748,29 @@ impl AppServer {
...
         // readiness event seems to be good enough™ for now.
 
+        println!("All sockets drained: {}", self.all_sockets_drained);
         // only poll if we drained all sockets before
         if self.all_sockets_drained {
             self.mio_poll.poll(&mut self.events, Some(timeout))?;
+
+            let queue_length = self.events.iter().peekable().count();
+
+            println!("queue length: {}", queue_length);
```
koraa, line 758: Debug message :)

rosenpass/src/protocol.rs
```diff
@@ -184,6 +195,89 @@ pub struct CryptoServer {
 
     // Tick handling
     pub peer_poll_off: usize,
+
```
koraa, line 198: I think it should be possible to reuse the code from the biscuit key?

rosenpass/src/protocol.rs
```diff
@@ -211,6 +305,8 @@ pub struct Peer {
     pub session: Option<Session>,
     pub handshake: Option<InitiatorHandshake>,
     pub initiation_requested: bool,
+    pub cookie_tau: CookieSecret,
```
koraa, line 308: I don't think storing the cookie in the peer is quite right, the cookie is bound to a specific handshake (through `sidi`), lets use it just for that…

Lets have it in the initiator handshake state…

rosenpass/src/protocol.rs
```diff
@@ -753,7 +856,122 @@ pub struct HandleMsgResult {
...
+    pub fn handle_msg_under_load(
+        &mut self,
+        rx_buf: &[u8],
+        tx_buf: &mut [u8],
+        peer: PeerPtr,
+        socket_addr: SocketAddr,
+    ) -> Result<HandleMsgResult> {
+        //Check cookie value
+        let mut ip_addr_port = match socket_addr.ip() {
+            IpAddr::V4(ipv4) => ipv4.octets().to_vec(),
```
koraa, line 876: I think it would be preferable to factor out IP-address specific functionality from protocol.rs; we want it to stay generic as opposed to relying on a specific network protocol…

koraa, line 876: There are a lot of memory allocations here, Rosenpass shares a goal with WireGuard, that network packages should not be able to trigger memory allocations…

rosenpass/src/protocol.rs
```diff
@@ -584,6 +685,8 @@ impl Peer {
             session: None,
             handshake: None,
             initiation_requested: false,
+            cookie_tau: CookieSecret::None,
+            last_sent_mac: None,
```
koraa, line 689: This information should already be present in the retransmission buffer…why don't we fetch it from there?

koraa (changes requested):
Wonderful work implementing this feature! I think it is very clear that we can get this work merged soon.

Two major things we should address though are:

- Not having IP-Address handling in protocol.rs (its supposed to be transport agnostic)
- Avoiding memory allocations

koraa:
@prabhpreet This PR makes you the first person outside of myself who dared to modify protocol.rs. Thank you for your contribution and well done!

koraa:
@prabhpreet So one thing I noticed is that your code does not trigger retransmission on CookieReply handling; it waits for the regular, scheduled retransmission. Is that correct?

How does WireGuard handle this?

rosenpass/src/app_server.rs
prabhpreet: Are peers configured with an IP address or an IP address range in RP? The intent was to find a peer index (or PeerPtr)

rosenpass/src/protocol.rs
prabhpreet: You're right, fixed

rosenpass/src/protocol.rs
prabhpreet: Have removed used of heap based memory allocations where possible

rosenpass/src/protocol.rs
prabhpreet: done

rosenpass/src/protocol.rs
prabhpreet: not sure I understand, could you elaborate?

papers/whitepaper.md
```diff
@@ -428,12 +428,89 @@ The responder code handling InitConf needs to deal with the biscuits and package
...
+
+When the responder is under load, it may choose to not process further handshake messages, but instead to respond with a cookie reply message (see Figure \ref{img:MessageTypes}). 
+
+The sender of the exchange then uses this cookie in order to resend the message and have it accepted the following time by the reciever. 
+
+For an initiator, Rosenpass ignores the message when under load.
+
+#### Cookie Reply Message
+
+The cookie reply message consists of the `sid` of the client under load, a random 24-byte bitstring `nonce` and encrypting `cookie_tau` into a `cookie` reply field which consists of the following (from the perspective of the cookie reply sender):
```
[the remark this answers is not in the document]
prabhpreet: called it `cookie_value` since there was already a `cookie_key` to encrypt cookies.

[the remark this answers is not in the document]
prabhpreet: done

papers/whitepaper.md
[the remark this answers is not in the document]
prabhpreet: called it `cookie_secret`

papers/whitepaper.md
prabhpreet: done

papers/whitepaper.md
prabhpreet: fixed

papers/whitepaper.md
prabhpreet: done

papers/whitepaper.md
prabhpreet: done

papers/whitepaper.md
prabhpreet: done, with session id specific correction you mentioned for the code

papers/whitepaper.md
prabhpreet: done

rosenpass/src/app_server.rs
koraa: Peers are sometimes not configured with any IP address at all. Consider the case of a client/server style setup: The client may be connected through mobile tethering, often with a changing IP-Address while the server may be on a home network, using upnp and dyndns.

In this scenario, the client looks up the server's ip address and sends an InitHello, the server will have no IP address for the client configured; the first connection will happen based on the client's InitHello message – the server can not send an InitHello because they don't know the client IP.

Once the handshake completes successfully (and only if it completes successfully!) the server will cache the client's IP address. Some time later, the client may switch IP addresses, in which case the IP address cached by the server will be incorrect. The server will keep trying to reconnect with an incorrect IP address, but with an incorrect IP.

The client can re-initiate the connection from their new IP address by sending an InitHello message and completing the handshake.

In this case the handshake must be performed even though the stored IP is a mismatch.

rosenpass/src/protocol.rs
koraa: The responder already issues a type of cookie called a "biscuit" at a later time in the handshake. The biscuit key functions similarly to the CookieKey, also needing to be replaced regularly, also storing a stale key to deal with the period after they key was replaced.

All that logic is already implemented in the BiscuitKey structure.

https://github.com/rosenpass/rosenpass/blob/184cff0e5ec1d67878801b62a91b4d21a679fb20/rosenpass/src/protocol.rs#L449

No need to redo all the work. Lets rename it "CookieKey" and use it for both the biscuit and cookie keys.

rosenpass/src/protocol.rs
koraa: Can you list the heap allocations you left in the code?

koraa:
> > So one thing I noticed is that your code does not trigger retransmission on CookieReply handling; it waits for the regular, scheduled retransmission. Is that correct? How does WireGuard handle this?
> 
> Yes that is correct. Wireguard does handle it the same way- the cookie reply does not cause retransmissions- the retranmissions happen as per Wireguard's retransmission logic, but with `mac.2` modified (equivalent to Rosenpass' `cookie` tag).

For reasons I am happy to discuss via chat but not publicly, can you:

- Make sure the CookieReply message is exactly as big as the InitHello message, by adding a padding field
- Trigger retransmission on CookieReply reception

rosenpass/src/protocol.rs
prabhpreet: None, removed all in the runtime code

rosenpass/src/protocol.rs
prabhpreet: done, named it `CookieStore`.

prabhpreet:
@koraa Resolved code review modifications except clients with unknown IP addresses initiating a connection with the server under load. Discussing in main thread for easier visibility:

> Peers are sometimes not configured with any IP address at all. Consider the case of a client/server style setup: The client may be connected through mobile tethering, often with a changing IP-Address while the server may be on a home network, using upnp and dyndns.
> 
> In this scenario, the client looks up the server's ip address and sends an InitHello, the server will have no IP address for the client configured; the first connection will happen based on the client's InitHello message – the server can not send an InitHello because they don't know the client IP.
> 
> Once the handshake completes successfully (and only if it completes successfully!) the server will cache the client's IP address. Some time later, the client may switch IP addresses, in which case the IP address cached by the server will be incorrect. The server will keep trying to reconnect with an incorrect IP address, but with an incorrect IP.
> 
> The client can re-initiate the connection from their new IP address by sending an InitHello message and completing the handshake.
> 
> In this case the handshake must be performed even though the stored IP is a mismatch.

The question then arises- how can a server find which peer it received a message from under load, without going through the computational intensive operations?

The peer is required to be known send a valid mac for the CookieReply message

rosenpass/src/app_server.rs
prabhpreet: commented concern on main thread

koraa:
> The question then arises- how can a server find which peer it received a message from under load, without going through the computational intensive operations?

It can't and let me point out that this knowledge is equivalent to having already performed some authentication, the entire purpose of rosenpass is to provide authentication.

The cookie mechanism is supposed to ensure that the initiator can not lie about their IP address to circumvent IP based blocking for instance. The cookie mechanism can not prevent DOS attacks in general.
 
> The peer is required to be known send a valid mac for the CookieReply message. Without a valid mac, the client (initiator) will simply discard the cookie reply message.

Yup, sending a valid Mac should be unproblematic?

papers/whitepaper.md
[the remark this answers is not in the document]
koraa: You can add the info:

1. Either in this issue
2. Or use a tool called xournalpp to produce a PDF with the places marked where changes need to be documented

koraa:
Can you give me push access to your fork?

prabhpreet:
> Yup, sending a valid Mac should be unproblematic?

But how is that possible, as a server, if you receive a packet without any knowledge of the IP address? As you said, as a server, we can't confidently map a peer from a packet without doing actual authentication. Under load, currently in the implementation,as a server we can simply best guess it in two cases:

1. If we already expect a client to have a IP+ Port
2. Use the IP+ Port used for the most recent handshake. 

A modification I can make to expand is allow any authenticated exchange to update the IP address. However, that won't solve the use case of when the server is under load for the following:

1. Creating a new connection (with a new InitHello message sent)
2. IP address changes after a connection has been established (roaming) and connecting with an entirely new handshake

I can't think of any other way- please let me know if there are solutions you have in mind!

papers/whitepaper.md
```diff
@@ -428,12 +428,86 @@ The responder code handling InitConf needs to deal with the biscuits and package
...
+#### Cookie Reply Message
+
+The cookie reply message is sent by the responder when under load. It consists of the `sidi` of the initiator, a random 24-byte bitstring `nonce` and encrypting `cookie_value` into a `cookie_encrypted` reply field which consists of the following:
+
+```pseudorust
+cookie_value = lhash("cookie-value", cookie_secret, initiator_host_info)[0..16]
+cookie_encrypted = XAEAD(lhash("cookie-key", spkm), nonce, cookie_value, mac_peer)
+```
+
+where `cookie_secret` is a secret variable that changes every two minutes to a random value. `initiator_host_info` is used to identify the initiator host, and is a concatenation of the initiator's IP address and UDP source port, with the bytes concatenated with big endian byte order. `cookie_value` is a truncated 16 byte value from the above hash operation. `mac_peer` is the `mac` field of the peer's handshake message to which message is the reply.
```
koraa, line 450: I just spent an hour trying to figure out if there is a unique encoding of the IP address and port, but I don't think there is. E.g. consider `fe80::` ipv6 link-local addresses which need an interface identifier.

Lets not specify that this is a concatenation of a IP address and UDP source port, instead lets define the host identification as implementation-specific and add a note cautioning implementers that values should be carefully chosen: Most implementations should take care to deal with IPv4 and IPv6 addresses, include the source port and make sure that special cases such as IPv6 link-local addresses are taken care of.

rosenpass/src/app_server.rs
```diff
@@ -584,6 +604,55 @@ impl AppServer {
...
+                    false
+                }
+            })
+            .ok_or(anyhow::anyhow!("Received message from unknown endpoint"))?
+            .0;
+
+        let mut len = 0;
+        let mut ip_addr_port = [0u8; 18];
+
+        match socket_addr.ip() {
```
koraa, line 638: This encoding has some collisions and omits some data; i.e. there exist many pairs of ipv6 and ipv4 addresses with the same binary representation.

Lets

1. Include the mio socket token (i.e. the socket index)
2. Represent ipv4 addresses as [ipv6-mapped-ipv4](https://www.ibm.com/docs/en/zos/2.2.0?topic=addresses-ipv4-mapped-ipv6)
3. Include the [ipv6 scope id](https://doc.rust-lang.org/std/net/struct.SocketAddrV6.html#method.scope_id)

Can we factor this out into a special "struct HostIdentification { ... }"?

koraa: Actually we already have a host identification: Endpoint::SocketBoundAddress.

Lets factor SocketBoundAddress into a struct `SocketBoundEndpoint` out.

try_recv can return the SocketBoundEndpoint and provide a nice `to_bytes` function as well as a `Display` implementation which can be used in the error handling code instead of manually printing the IP address there

rosenpass/src/app_server.rs
```diff
@@ -584,6 +604,55 @@ impl AppServer {
...
+                ip_addr_port[0..16].copy_from_slice(&ipv6.octets());
+                len += 16;
+            }
+        };
+
+        ip_addr_port[len..len + 2].copy_from_slice(&socket_addr.port().to_be_bytes());
+        len += 2;
+
+        self.crypt
+            .handle_msg_under_load(&rx[..len], &mut *tx, PeerPtr(index), &ip_addr_port[..len])
```
koraa, line 653: Can we just have one handle_msg function which handles messages under load? This would also greatly simplify the `ReceivedMessage` case in `event_loop`

koraa (changes requested):

rosenpass/src/protocol.rs
```diff
@@ -754,7 +852,127 @@ pub struct HandleMsgResult {
...
-    /// Respond to an incoming message
+    /// Process a message under load
+    /// This is one of the main entry point for the protocol.
+    /// Keeps track of messages processed, and qualifies messages using
+    /// cookie based DoS mitigation. Dispatches message for further processing
+    /// to `process_msg` handler if cookie is valid, otherwise sends a cookie reply
+    /// message for sender to process and verify for messages part of the handshake phase
+    /// (i.e. InitHello, InitConf messages only). Bails on messages sent by responder and
+    /// non-handshake messages.
+    pub fn handle_msg_under_load(
```
koraa, line 863: Can we avoid the code duplication here and merge handle_msg and handle_msg_under_load?

rosenpass/src/protocol.rs
```diff
@@ -754,7 +852,127 @@ pub struct HandleMsgResult {
...
+        &mut self,
+        rx_buf: &[u8],
+        tx_buf: &mut [u8],
+        peer: PeerPtr,
+        ip_addr_port: &[u8],
+    ) -> Result<HandleMsgResult> {
+        let mut cookie_value = [0u8; 16];
+        match self.cookie_secret.lifecycle(&self) {
```
koraa, line 871: Can we avoid the code duplication and use the same code as in active_biscuit_key?

rosenpass/src/protocol.rs
```diff
@@ -276,6 +298,9 @@ pub struct InitiatorHandshake {
     pub tx_count: usize,
     pub tx_len: usize,
     pub tx_buf: MsgBuf,
+
+    // Cookie storage for retransmission, expires PEER_COOKIE_VALUE_EPOCH seconds after creation
+    pub cookie_value: CookieStore<COOKIE_VALUE_LEN>,
```
koraa, line 303: This – like the biscuit key – should store two keys; a stale one which is still good for decrypting and a new one for encrypting and decrypting.

The best way to achieve this is probably to factor `[CookieStore<KEY_LEN>; 2]` out into a new structure; something like `CookeRotation` which handles erasure of stale keys, encryption, and decription operations.

koraa (approved):

koraa (changes requested):
This is nicely coming together; a few issues are left to be addressed:

- Code duplication in protocol.rs
- Unique serialization of IP addresses
- Immediately responding to cookie replys
- Adding comments to the graphics to indicate what needs updating

koraa:
Can I just restate again, that I am pretty impressed by the work you did here? This is a major update to the protocol handling code and despite some minor issues like code duplication it is really looking good!

koraa:
@prabhpreet Would you be interested in approaching integration tests in follow up work? We do have budget for it, but I did not want to issue a bounty, since this is something that should be handled by someone more experienced with the code…

rosenpass/src/protocol.rs
koraa: I think we can pass the load condition information as a parameter to handle_msg…

rosenpass/src/protocol.rs
```diff
@@ -1735,10 +1997,75 @@ impl CryptoServer {
 
         Ok(hs.peer())
     }
+
+    pub fn handle_cookie_reply(&mut self, cr: CookieReply<&[u8]>) -> Result<PeerPtr> {
```
koraa, line 2001: Lets avoid the super-deeply nested code here; there is a lot of rosenpass code demonstrating how `if ERROR { bail/return }` or `let VAR = match value {
 ERROR => bail,
 GOOD => value,
}` can be used to avoid unnecessarily deep indendation

koraa (changes requested):

koraa:
> > Yup, sending a valid Mac should be unproblematic?
> 
> But how is that possible, as a server, if you receive a packet without any knowledge of the IP address a peer is supposed to have?

There are two mac values; currently `mac` and `cookie`.

- `mac` is always generated and guards the packed against modification unless the recipient static key is known
- `cookie` is generated using the cookie value; this value is __only__ checked under load. The cookie mac is generated using what we now call `cookie_value`. The cookie value is derived from the __sender's__ host identification even if the sender is a fake, not from the peer's true IP address or port. Since we use the sender's host identification, this information is implicitly sent along with every IP package.

> As you said, as a server, we can't confidently map a peer from a packet without doing actual authentication. Under load, currently in the implementation,as a server we can simply best guess it in two cases:
> 
>     1. If we already expect a client to have a IP+ Port
> 
>     2. Use the IP+ Port used for the most recent handshake
> 
> 
> A modification I can make to expand is allow any authenticated exchange to update the IP address.

The point of the cookie mechanism is to prevent senders from lying about their IP address, i.e. prevent an __IP spoofing__ attack.  The peer's true IP does not factor into it.

Consider this case: An attacker performs a DOS attack, the server admin notes this and blocks the attacker's IP in the firewall. To adapt to this countermeasure, the attacker spoofs their IP, setting a random sender ip in their packet. Since the attacker's ISP has bad admins, they do not prohibit this IP spoofing attack.

Due to the load, the Rosenpass cookie mechanism kicks in, forcing clients to prove that they possess the IP they pretend to have, this automatically leads to an early rejection of most of the attacker's packages.

> However, that won't solve the use case of when the server is under load for the following:
> 
>     1. Creating a new connection (with a new InitHello message sent
> 
>     2. IP address changes after a connection has been established (roaming) and connecting with an entirely new handshake
> 
> 
> I can't think of any other way- please let me know if there are solutions you have in mind!


I think the issue arises from a misunderstanding of the purpose of the cookie reply mechanism, let me know if the above comment clarifies the questions.

> > Can you give me push access to your fork?
> 
> You should have recieved an invite at @koraa , please let me know if you didn't.

Thanks.

koraa:
@prabhpreet I am trying to merge the code with main right now and I just noticed you added cookie reply handling for InitConf. InitConf uses no asymmetric crypto operations, so we do not have to handle cookies for init conf…

koraa:
@prabhpreet I merged main into this for you. Lets fix the errors and get this merged…

koraa:
> > @prabhpreet I am trying to merge the code with main right now and I just noticed you added cookie reply handling for InitConf. InitConf uses no asymmetric crypto operations, so we do not have to handle cookies for init conf…
> 
> @koraa This looks like a change to the behaviour we previously discussed (mentioned in the changes in whitepaper)
> 
> To confirm, the behaviour now will be that a responder will just process InitConf as is, even if under load and not reply with a cookie reply message? So cookie value will only check for valid cookies for incoming InitHello messages for a responder? And we will ignore processing any other message types?

I think that is all right, yes!

PR #143, opened 2023-11-04
refactor rp to workspace

wucke13:
We have multiple endeavors on the way which require the creation of multiple crates. IMHO, it is best to organize these crates in a big mono-repo, using a Cargo Workspace.


koraa (approved):

koraa:
@wucke13 Did you make sure the ./rp script continues to function?

koraa:
@wucke13 I think I was able to tell naersk that this is a workspace; could you please make sure the `rp` script continues to function in `nix develop` before merge?

PR #154, opened 2023-11-19
Add an initial set of fuzzing targets

pcwizz:
These targets can be used with rust nightly and cargo-fuzz to fuzz several bits of Rosenpass's API. Fuzzing is an automated way of exploring code paths that may not be hit in unit tests or normal operation. For example the `handle_msg` target exposed the DoS condition fixed in 0.2.1.

The other targets focus on the FFI with libsodium and liboqs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "2"
 
 members = [
     "rosenpass",
+    "fuzz",
 ]
 
 [workspace.metadata.release]
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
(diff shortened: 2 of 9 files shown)


koraa:
Github is being slow and I can not push; @pcwizz Could you check the following modifications:

[0001-chore-Prepare-fuzzing-PR-for-merge.patch.txt](https://github.com/rosenpass/rosenpass/files/13411995/0001-chore-Prepare-fuzzing-PR-for-merge.patch.txt)

koraa:
@pcwizz Could you please also mark your commit as `feat(fuzzing): `

koraa:
Merged successfully through #162 :fireworks: :fireworks: :fireworks: 

@pcwizz Thank you for this contribution!!

PR #159, opened 2023-11-20
doc: Add bibliography to the manual page

### LINKED ISSUE #153: Add reference to paper to the man page
You should add a reference to your publication to the man page.  You can use the `.Rs` request to generate bibliography entries, see mdoc(7) for details.

emilengler:
Fixes #153

cc @clausecker


clausecker:
I'd use `%T` instead of `%B` as it's a paper, not a book.  You could also use `%R` perhaps, but it doesn't look as great.

emilengler:
Done.

koraa (approved):

Issue #121, opened 2023-09-13
ASCII Armored Keys


lorenzleutgeb:
The Rosenpass key files (commonly `pqpk` and `pqsk`, generated by `rosenpass gen-keys`) are binary files. In some contexts, binary files are cumbersome/brittle to handle. Please consider supporting [ASCII armoring](https://en.wikipedia.org/wiki/Binary-to-text_encoding#ASCII_armor), similar to the [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) format used in other software.

What I did to work around my issues with binary files was to simply call [`base64`](https://www.gnu.org/software/coreutils/manual/coreutils.html#base64-invocation) to encode/decode the files. It should be fairly simple to allow Rosenpass to read/write base64 encoded keys.

A related issue also came up in https://github.com/rosenpass/rosenpass/issues/64#issuecomment-1642064942

emilengler:
I'm okay with accepting a change to that.
The issue I only see with it, would be how to handle backwards compatibility in that case ...

emilengler:
The `pem-rfc7468` seems to be suitable to our use case.
I only fear that such a change would lead to new side-channel attack. After all, a secret key will be accessed by yet another function.

stv0g:
I am against using ASCII armored keys for two reasons:
- Classic McEliece keys are rather huge (> 2MB?). So a ASCII-printable encoding just increases the keys even more because of encoding inefficiencies.
- ASCII armored keys seems to encourage people attempting copy&pasting keys via clip-board or sending they via instant messaging / mail. Given the key sizes, I think this is not a really suitable approach, because which clipboard buffer can handle 2 MB contents or what happens if you send a 2 MB XMPP message?

Using binary files would encourage people to exchange the key contents as files and not some other means.

emilengler:
The size of classic mceliece pubkeys is definitely a good point.

lorenzleutgeb:
> * Classic McEliece keys are rather huge (> 2MB?). So a ASCII-printable encoding just increases the keys even more because of encoding inefficiencies.

Yes, I am aware that base64 encoding increases file size and that large keys are problematic in general. I don't see how any encoding would magically decrease the size of key files while preserving their entropy. The point is usability, and is sufficiently motivated by key armoring being supported in widely known systems like GPG, SSH and TLS.

<table>
[...]

PR #161, opened 2023-11-22
encode local keys in base64

aparcar:
Wireguard keys are always handled base64 encoded which makes it easier to share and move around. At least the local public key would be shared one way or another.

This breaks existing setups, however I wonder if long term a switch to base64 encoded keys would be best. I couldn't find any similar comment in the issues but maybe I missed something, if so please close this.

### DIFF
--- a/rosenpass/src/cli.rs
+++ b/rosenpass/src/cli.rs
@@ -1,15 +1,17 @@
 use anyhow::{bail, ensure};
 use clap::Parser;
+use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use crate::app_server;
 use crate::app_server::AppServer;
-use crate::util::{LoadValue, LoadValueB64};
+use crate::util::LoadValueB64;
 use crate::{
     // app_server::{AppServer, LoadValue, LoadValueB64},
     coloring::Secret,
     pqkem::{StaticKEM, KEM},
     protocol::{SPk, SSk, SymKey},
+    util::{b64_writer, fopen_w},
 };
 
 use super::config;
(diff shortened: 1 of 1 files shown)


koraa:
@aparcar Thank you for submitting this pull request. I think this works, but I am doubtful that this is a good idea. The public key is a  >100kB file; sharing this anywhere online via copy-paste is hard to conceive of…

koraa:
I just tried pasting a key with this patch, but either github or firefox refuse…

aparcar:
I'm mostly thinking that both the Wireguard key and the Wireguard PSK are stored base64 encoded so it would add some consistency for file handling. To have everything in a single file it would also be easier to handle if it's base64 encoded.

koraa:
> I'm mostly thinking that both the Wireguard key and the Wireguard PSK are stored base64 encoded so it would add some consistency for file handling. To have everything in a single file it would also be easier to handle if it's base64 encoded.

How would you put everything in a single file?

./cc @wucke13 @emilengler @stv0g What to you think about this?

emilengler:
I'm not a fan of having everything in a file. This would require interfacing applications to implement a parsing logic. While it is only a small overhead, I believe that it is an unnecessary one.

koraa:
I would not mind ascii armor; there are situations where that could be useful. I would mind text files that explode most people's text editors.

In any case, using ascii armor should remain an optional off-by-default feature and we should not allow any data below or above the ascii armor…

wucke13:
So far, we did not see conclusive argumentation to move away from binary files. Most notably however, the use of PEM for keys sounds interesting. Feel free to continue the discussion in favour/against in https://github.com/rosenpass/rosenpass/issues/121 up until then.

PR #171, opened 2023-11-28
chore: Move hashing functions into sodium/ciphers crate

koraa:
This finishes the last step of removing sodium.rs from the rosenpass crate itself and also removes the NOTHING and NONCE0 constants.

Hashing functions now use destination parameters;
rosenpass_constant_time::xor now does too.


koraa:
@wucke13 Contains some nix changes.

Will be merged tomorrow.

PR #174, opened 2023-11-28
Remove unwrap from fuzz targets that return errors

pcwizz:
When fuzzing we are interested in what happens inside the target function not necessarily what it returns. Functions returning errors with bogus input in generally desired behaviour.


koraa:
Good point!

PR #181, opened 2023-11-30
Improve the safety of the secret storage mechanism

koraa:
This exposes the libsodium allocation API as an allocator and uses that together with allocator-api2's Box and Vec to implement the secret pool.

This brings the number of unsafe blocks needed to implement this down to two from six.


koraa:
@pcwizz How would you feel about also adding fuzzing for the libsodium based allocator?

PR #195, opened 2023-12-03
Enhancement/secret allocation

### LINKED ISSUE #179: Implement secret memory in pure rust
Currently we are using libsodium's secret memory features through `sodium_malloc`. Sodium provides guard pages and uses mlock to avoid the secret memory area being swapped to disk.

We should create a new backend in rosenpass-secret-memory that:

- Uses [memfd_secret(2)](https://man.archlinux.org/man/memfd_secret.2.en) and guard pages using [mmap(2)](https://man.archlinux.org/man/mmap.2)
- If memfd_secret is unavailable, we should implement the same strategy as sodium guard pages and [mlock(2)](https://man.archlinux.org/man/mlock.2)
- Failing that we should continue to provide a sodium based secret memory backend
- Or allow users to specify that a dummy implementation using plain memory allocation should be used

Shylock-Hg:
Close #179 
/claim #179 

This PR use rust OnceLock to dispatch this new allocation to:
1. memfd_secret if supported
2. fallback to memsec


sodium/build.rs
```diff
@@ -0,0 +1,31 @@
...
+    };
+    if fd != -1 || if let Some(os_err) = Error::last_os_error().raw_os_error() { os_err != 38 } else { true } {
+        // with memfd_secret support
+        println!("cargo:rustc-cfg=with_memfd_secret");
+    }
+    if fd != -1 {
+        unsafe {
+            libc::close(fd);
+        }
+    }
```
wucke13, line 19: This is a leaky abstraction, that only produces correct results when `buildPlatform == hostPlatform`. This is not good, Rust has quite good cross-compilation support, and once we get rid of our C dependencies it should be trivial to cross compile Rosenpass. If I understand this code correctly, it breaks that, because it uses syscalls on the buildPlatform to detect which features may be available on the hostPlatform.

Cargo.lock
```diff
@@ -1181,6 +1181,7 @@ version = "0.1.0"
 dependencies = [
  "allocator-api2",
  "anyhow",
+ "libc",
```
wucke13, line 1184: Not adding a lot of dependencies, excellent!

wucke13:
I erroneously referred to this very PR up above, but I meant to refer to #179 where I left some comments with regard to the implementation of this feature. Please check them once and comment there.

koraa:
> > I erroneously referred to this very PR up above, but I meant to refer to #179 where I left some comments with regard to the implementation of this feature. Please check them once and comment there.
> 
> I checked it. But seems I can't answer this, use `memfd_secret` is the requirement of this issue, I guess your team have got agree with this before.

@wucke13 Is one of the main devs of this crate. And he is right – if memsec exists we should use it. Can you check how good the memsec crate's code is internally?

@Shylock-Hg Can you first of all please move your code into the -secret-memory crate? The -sodium crate is for C bindings to the sodium library. The memsec crate does not implement memfd_secret, this remains a better mechanism for secret storgae on linux, but memsec can serve as the fallback implementation using just memlock.

sodium/build.rs
koraa: If detecting syscalls is not possible at compile time, we should do it at runtime!

koraa:
> > > > I erroneously referred to this very PR up above, but I meant to refer to #179 where I left some comments with regard to the implementation of this feature. Please check them once and comment there.
> > > 
> > > 
> > > I checked it. But seems I can't answer this, use `memfd_secret` is the requirement of this issue, I guess your team have got agree with this before.
> > 
> > 
> > @wucke13 Is one of the main devs of this crate. And he is right – if memsec exists we should use it. Can you check how good the memsec crate's code is internally?
> > @Shylock-Hg Can you first of all please move your code into the -secret-memory crate? The -sodium crate is for C bindings to the sodium library. The memsec crate does not implement memfd_secret, this remains a better mechanism for secret storgae on linux, but memsec can serve as the fallback implementation using just memlock.
> 
> Ok, so according to discuss, IMO current solution is :
> 
>     1. Move related code to crate `secret-memory` to complete this issue
> 
>     2. still need `memfd_secret` if supported
> 
>     3. fallback to the crate `memsec` finally
>        All right? @wucke13 @koraa

Yes! Please go ahead with the implementation. Thank you for your contribution!

wucke13:
@Shylock-Hg We took some three hours to review your PR, and found major flaws in it. It can not work the way it its (just using the same fd over and over again). The `memfd_secret` code is plain broken. It seems like the `memfd_secret` related code is completely untested. Because of that we are aborting this attempt.

secret-memory/src/allocator.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+#[cfg(unix)]
+fn get_support_memfd_secret() -> bool {
+    static SUPPORT_MEMFD_SECRET: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+    *SUPPORT_MEMFD_SECRET.get_or_init(|| {
+        let fd  = unsafe {
+            libc::syscall(libc::SYS_memfd_secret, 0) as i32
+        };
+        if fd != -1 || if let Some(os_err) = Error::last_os_error().raw_os_error() { os_err != 38 } else { true } {
```
koraa, line 81: This code is weird as hell; Error::last_os_error is the incorrect abstraction level (should use libc's builtin errno support).

errno is never being cleared, so this could fail spuriously based on the last invocation

secret-memory/src/allocator.rs
```diff
@@ -0,0 +1,239 @@
...
+                with offset {off} from the requested alignment. Memsec always allocates values \
+                at the end of a memory page for security reasons, custom alignments are not supported. \
+                You could try allocating an oversized value.");
+            return Err(AllocError);
+        }
+
+        Ok(ptr)
+    }
+
+    fn do_deallocate(&self, ptr: NonNull<u8>, _layout: Layout) {
```
koraa, line 124: This function should be marked unsafe. WHy not just implement the Allocator trait here as well?

secret-memory/src/allocator.rs
```diff
@@ -0,0 +1,239 @@
...
+
+impl SecretAlloc {
+    pub fn new() -> Self {
+        let fd  = unsafe {
+            libc::syscall(libc::SYS_memfd_secret, 0) as i32
+        };
+        if fd == -1 {
+            log::error!(
+                "Create secret file descriptor failed {}.",
+                std::io::Error::last_os_error()
```
koraa, line 145: Again – wrong abstraction level. Use libc's errno support.

secret-memory/src/allocator.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    fn do_allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
+        if self.fd == -1 {
+            log::error!(
+                "Create secret file descriptor failed."
+            );
+            return Err(AllocError);
+        }
+        let ptr = unsafe {
+            let ret = libc::mmap(null_mut::<libc::c_void>(), layout.size(), libc::PROT_READ | libc::PROT_WRITE, libc::MAP_LOCKED, self.fd, 0);
```
koraa, line 161: This is missing the MAP_PRIVATE or MAP_SHARED flag, so this will always fail:

> The flags argument determines whether updates to the mapping are
>       visible to other processes mapping the same region, and whether updates
>       are carried through to the underlying file.  This behavior is
>       determined by including exactly one of the following values in flags:

secret-memory/src/allocator.rs
```diff
@@ -0,0 +1,239 @@
...
+        unsafe {
+            memsec::free(ptr)
+        }
+    }
+}
+
+
+#[derive(Clone)]
+struct SecretAlloc {
+    fd : i32,
```
koraa, line 134: This will always allocate *the same secret* so this is not a functioning allocator.

koraa (commented):
I am very sorry to be so blunt here, but this PR exemplifies why coding unsafe code can be extremely dangerous. You might have used the correct functions, but you omitted reading their documentation leading to insecure, erroneous code.

Study the manpages `mmap(2)`, `ftruncate(2)` and `memfd_secret(2)` and read them carefully. Coding by "I think these are the right functions and it does not crash" is inacceptable for unsafe rust code.

PR #196, opened 2023-12-03
fix: remove OSFONTDIR var from whitepaper build

### LINKED ISSUE #65: Remove unnecessary OSFONTDIR env variable
https://github.com/rosenpass/rosenpass/blob/d5b2a9414fd6f5f3bf550fddb79e5c684ba461cf/flake.nix#L268

wucke13:
Fixes #65. I checked with `pdffonts` that the whitepaper still has all fonts embedded.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -291,7 +291,6 @@
               ];
               buildPhase = ''
                 export HOME=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


TeXhackse (approved):
Doesn't differ from what we had discussed months ago, so it's fine.

PR #198, opened 2023-12-09
feat: First version of broker based WireGuard PSK interface

koraa:
This allows us to run with minimal priviledges in the Rosenpass process itself.

The PR is currently a draft because we need to improve the code quality to merge this…

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,7 +179,7 @@ dependencies = [
  "regex",
  "rustc-hash",
  "shlex",
- "syn",
+ "syn 2.0.39",
  "which",
 ]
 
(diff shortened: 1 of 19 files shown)


aparcar:
While trying to get things working I faced some questions:
* This commit introduces a hard dependency on Linux while it would previously work on i.e. MacOS, too. Could we make the `wireguard-broker` optional?
* Could you please provide me with a command on how to start this? For the `broker-privileged` exists no usage message and I'm a bit in the dark how to combine it with the `broker-socket-handler`.
* Could we define a message format to be run independently of Wireguard? I'd like to use Rosenpass as a *magic bullet* to initiate the exchange of key material via a socket message, if possible.

aparcar:
Is the socket API for the broker documented? I'm wondering if I can feed PSKs into wireguard from another service.

koraa:
> Is the socket API for the broker documented? I'm wondering if I can feed PSKs into wireguard from another service.

Not yet, can you add a ticket and specify your use case specifically?

PR #224, opened 2023-12-27
feat: Migrate to memsec

koraa:


koraa:
Looks like the fuzzer might actually have found something?  https://github.com/rosenpass/rosenpass/actions/runs/7339888800/job/19984940094?pr=224

thaodt:
> Looks like the fuzzer might actually have found something? https://github.com/rosenpass/rosenpass/actions/runs/7339888800/job/19984940094?pr=224

interesting... could you please share your experiment with this fuzz test? what will you do in this case? how do you debug and fix it? when i look at the error, it only displayed the fuzz fn `fuzz_handle_msg`, I expected it could show us which line number caused crash..

wucke13:
I can reproduce the CI failure with this:
```console
cargo fuzz run fuzz_mceliece_encaps -- -max_total_time=60
```

wucke13:
Ok, I found the first problem that causes fuzz failure:

https://github.com/rosenpass/rosenpass/blob/a477d2d791e4b542b6469e1bf1e38f4fafa9fea3/fuzz/fuzz_targets/mceliece_encaps.rs fails, because `StaticKem::encaps` is passed an empty slice as `sk`. In particular, [this `assert_eq!`](https://github.com/rosenpass/rosenpass/blob/a477d2d791e4b542b6469e1bf1e38f4fafa9fea3/oqs/src/kem_macro.rs#L30) is the culprit. Now I'm wondering, if we should either 

- statically ensure that a slice of the correct length is passed (by making the arg a `&[u8; Self::SK_LEN]`
- handle the error dynamically using Result, instead of panicking on wrong buffer size
- limit the fuzzer to only fuzz with inputs of correct length

Anyhow I added a test case for the first of the two failures to this branch.

wucke13:
The second error I do not understand yet. It appears the failure occurs for an _empty input_, at least the unit file is empty and this output is generated by `cargo fuzz`:

```console
Failing input:

	fuzz/artifacts/fuzz_handle_msg/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709

Output of `std::fmt::Debug`:

	[]
```

Despite that, my minimal unit test that provides empty output does __not__ fail: 

```rust
    #[test]
    fn fuzz_handle_msg_error_2023_12_28() {
        sodium_init().unwrap();

        let sk = Secret::from_slice(&[0; 13568]);
        let pk = Secret::from_slice(&[0; 524160]);

        let mut cs = CryptoServer::new(sk, pk);
        let mut tx_buf = [0; 10240];

        let _ = cs.handle_msg(&[], &mut tx_buf);
    }
```

thaodt:
> The second error I do not understand yet. It appears the failure occurs for an _empty input_, at least the unit file is empty and this output is generated by `cargo fuzz`:
> 
> ```
> Failing input:
> 
> 	fuzz/artifacts/fuzz_handle_msg/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709
> 
> Output of `std::fmt::Debug`:
> 
> 	[]
> ```
> 
> Despite that, my minimal unit test that provides empty output does **not** fail:
> 
> ```rust
>     #[test]
>     fn fuzz_handle_msg_error_2023_12_28() {
>         sodium_init().unwrap();
> 
>         let sk = Secret::from_slice(&[0; 13568]);
>         let pk = Secret::from_slice(&[0; 524160]);
> 
>         let mut cs = CryptoServer::new(sk, pk);
>         let mut tx_buf = [0; 10240];
> 
>         let _ = cs.handle_msg(&[], &mut tx_buf);
>     }
> ```

same here. I tried to replicate and really didn't understand the failure caused by an _empty input_.

wucke13:
@thaodt may it bee that this really is a stack overflow? Does the fuzzing framework allocate the values on stack? We had trouble with stack overflows before, always when touching Classic McEliece. I only tried the env var which changes the stack size for new threads, the main thread still has the rust default. We should try changing that as well.

wucke13:
In particular, the fuzzer function basically asks for a stack overflow:

```rust
fuzz_target!(|input: [u8; StaticKem::PK_LEN]| {
    let mut ciphertext = [0u8; 188];
    let mut shared_secret = [0u8; 32];

    // We expect errors while fuzzing therefore we do not check the result.
    let _ = StaticKem::encaps(&mut shared_secret, &mut ciphertext, &input);
});
```

IIRC, `StaticKem::PK_LEN` is somewhere around 500 kilobytes. So only putting the function call's arguments (which are passed by value (!) to the stack already adds 0.5 megabytes of stack consumption.

koraa:
> Ok, I found the first problem that causes fuzz failure:
> 
> https://github.com/rosenpass/rosenpass/blob/a477d2d791e4b542b6469e1bf1e38f4fafa9fea3/fuzz/fuzz_targets/mceliece_encaps.rs fails, because `StaticKem::encaps` is passed an empty slice as `sk`. In particular, [this `assert_eq!`](https://github.com/rosenpass/rosenpass/blob/a477d2d791e4b542b6469e1bf1e38f4fafa9fea3/oqs/src/kem_macro.rs#L30) is the culprit. Now I'm wondering, if we should either
> 
>     * statically ensure that a slice of the correct length is passed (by making the arg a `&[u8; Self::SK_LEN]`
> 
>     * handle the error dynamically using Result, instead of panicking on wrong buffer size
> 
>     * limit the fuzzer to only fuzz with inputs of correct length
> 
> 
> Anyhow I added a test case for the first of the two failures to this branch.

I would say this is correct behavior, so the fuzzing code needs adjusting…

koraa:
I am knee-deep into GDB at this point due to the handle_msg failure.

I can see calls such as the following:

```
#0  GetTLSFakeStack () at /rustc/llvm/src/llvm-project/compiler-rt/lib/asan/asan_fake_stack.cpp:176
#1  GetFakeStackFast () at /rustc/llvm/src/llvm-project/compiler-rt/lib/asan/asan_fake_stack.cpp:193
#2  OnMalloc () at /rustc/llvm/src/llvm-project/compiler-rt/lib/asan/asan_fake_stack.cpp:207
#3  __asan_stack_malloc_6 () at /rustc/llvm/src/llvm-project/compiler-rt/lib/asan/asan_fake_stack.cpp:261
#4  0x0000555556286de0 in rosenpass::protocol::CryptoServer::handle_msg (self=0x3e80, rx_buf=..., tx_buf=...)
    at rosenpass/src/protocol.rs:785
```

I think somehow the address sanitizer or the LeakSanitizer might be interfering with memsec.

grim_1703857943

koraa:
Issue found and fixed. Check commit log.

koraa:
@wucke13 Please review and merge

wucke13:
@koraa Impressive work. This looks mostly good to me. I have two remaining suggestions:

1. Shall we close https://github.com/quininer/memsec/issues/14 again (with an explanation), now that we know this was an issue with address sanitizer + lazy_static?
2. There is a bunch of small clippy lints, I would like to push one additional commit that fixes them.

Just give me thumbs if you agree, or comment if you have other suggestions.

thaodt:
> 1. (with an explanation), now that we know this was an issue with address sanitizer + lazy_static?
can you elaborate more on it @wucke13 ?

PR #258, opened 2024-03-07
ci: enable cargo bench again

aparcar:
It only takes a few seconds to run, enable it.

### DIFF
--- a/.github/workflows/qc.yaml
+++ b/.github/workflows/qc.yaml
@@ -51,7 +51,7 @@ jobs:
         # liboqs requires quite a lot of stack memory, thus we adjust
         # the default stack size picked for new threads (which is used
         # by `cargo test`) to be _big enough_. Setting it to 8 MiB
-      - run: RUST_MIN_STACK=8388608 cargo bench --no-run --workspace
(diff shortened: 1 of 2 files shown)


aparcar:
@koraa please have a look why this fails

```
Benchmarking handshake: Warming up for 3.0000 s
thread 'main' panicked at rosenpass/benches/handshake.rs:77:54:
called `Result::unwrap()` on an `Err` value: Rejecting biscuit: Outdated biscuit number
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

koraa:
@aparcar I overlooked that memsec::memcmp is doing a lexicographical comparison which is also know as a big endian comparison.

Would you care to update rosenpass_constant_time::compare to perform a right-to-left comparison, i.e. a little endian comparison? My brain is mush right now…

You could start from https://github.com/quininer/memsec/blob/bbc647967ff6d20d6dccf1c85f5d9037fcadd3b0/src/lib.rs#L32

aparcar:
I tried to fix it via https://github.com/rosenpass/rosenpass/pull/262 but I'm not entirely sure if that's the clean way.

koraa (approved):

PR #259, opened 2024-03-09
fix: Fix crate vulnerabilities

rex4539:
https://rustsec.org/advisories/RUSTSEC-2024-0019
https://rustsec.org/advisories/RUSTSEC-2024-0006

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -868,9 +868,9 @@ dependencies = [
 
 [[package]]
 name = "mio"
-version = "0.8.9"
+version = "0.8.11"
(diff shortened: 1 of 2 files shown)


emilengler:
@wucke13 Do you know about an automated way to check for Rust vulnerabilities?

emilengler:
This by the way probably does not require a security release, because as far as I understand it, the issue is only present on Windows, which we do not officially support anyway.

koraa:
@emilengler You know you are allowed to merge these PRs :)
@rex4539 Thank you for the contribution!

PR #263, opened 2024-03-12
bench: exclude rosenpass-fuzzing

aparcar:
This stops fuzzing to run which takes forever and breaks the CI.


koraa (changes requested):
Lets specifically disable the fuzzing tests, not disable most benchmarks…

Issue #260, opened 2024-03-09
Enforce proper permissions in `store_secret`


emilengler:
I recently came across the following piece of code in the secret-memory crate:
```rust
impl<const N: usize> StoreSecret for Secret<N> {
    type Error = anyhow::Error;

    fn store_secret<P: AsRef<Path>>(&self, path: P) -> anyhow::Result<()> {
        std::fs::write(path, self.secret())?;
        Ok(())
    }
}
```
As far as I understand this, it does not ensure that the file has to proper permissions.
This is not acceptable, a secret must never be written into an insecure file.

PR #265, opened 2024-03-13
proper permission for secrets aka 0o600

aparcar:
When creating secret keys or use the out file feature, the material shouldn't be readble to everyone by default.

Fix: #260

### DIFF
--- a/util/src/file.rs
+++ b/util/src/file.rs
@@ -1,17 +1,24 @@
 use anyhow::ensure;
 use std::fs::File;
 use std::io::Read;
+use std::os::unix::fs::OpenOptionsExt;
 use std::result::Result;
 use std::{fs::OpenOptions, path::Path};
 
+pub enum Visibility {
+    Public,
+    Secret,
+}
+
 /// Open a file writable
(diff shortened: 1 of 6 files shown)


util/src/file.rs
```diff
@@ -1,17 +1,18 @@
...
 
 /// Open a file writable
-pub fn fopen_w<P: AsRef<Path>>(path: P) -> std::io::Result<File> {
-    OpenOptions::new()
-        .read(false)
-        .write(true)
-        .create(true)
-        .truncate(true)
-        .open(path)
+pub fn fopen_w<P: AsRef<Path>>(path: P, secret: bool) -> std::io::Result<File> {
```
koraa, line 9: Can we refactor to fopen_w_public and fopen_w_secret? Or use a `enum Visibility { Public, Secret }`?

koraa (approved):
This is great, thank you!
In the name of developer experience we might avoid a boolean and instead use something more descriptive.

aparcar:
We can just use the permission directly

util/src/file.rs
aparcar: Reworked with an `enum`.

koraa (approved):

PR #266, opened 2024-03-13
update to liboqs 0.9.1

aparcar:
The release updated the Classic McEliece implemented and since we use it, might be a good idea to update the lib.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,11 +173,11 @@ checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
 name = "bindgen"
-version = "0.65.1"
+version = "0.68.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cfdf7b466f9a4903edc73f95d6d2bcd5baf8ae620638762244d3f60143643cc5"
+checksum = "726e4313eb6ec35d2730258ad4e15b547ee75d6afaa1361a922e78e59b7d8078"
 dependencies = [
- "bitflags 1.3.2",
+ "bitflags 2.4.1",
  "cexpr",
  "clang-sys",
  "lazy_static",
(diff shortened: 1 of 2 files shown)


koraa:
This is fascinating

```
thread 'fuzz_handle_msg' panicked at /home/runner/work/rosenpass/rosenpass/secret-memory/src/secret.rs:152:31:
source slice length (13568) does not match destination slice length (13608)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Does the fuzzer crash reproduce when you run locally?

koraa:
Oh by the way, it is possible that the updated OQS actually has new versions of McEliece & Kyber, in which case this crash would make sense

aparcar:
> Oh by the way, it is possible that the updated OQS actually has new versions of McEliece & Kyber, in which case this crash would make sense

How so?

aparcar:
In other words, does this mean with a bump all deployed setups would break?

prabhpreet:
Looking into the cargo fuzz failure, will add any findings here

aparcar:
Great thanks

prabhpreet:
The secret key size and the cipher text length have changed in the McEliece KEM we are using. 

https://github.com/open-quantum-safe/liboqs/blob/7c3a0e9aa7f9568e4dcafaf908ff8aa0008f0b71/src/kem/classic_mceliece/kem_classic_mceliece.h#L32-L33

So the slice size turns out to be smaller when copying the slice into Secret:

https://github.com/aparcar/rosenpass/blob/5c84df41454a36bf3a0241583273e6f694414aa0/fuzz/fuzz_targets/handle_msg.rs#L10

I've added some updates to use constants instead of manually specifying slice sizes in these tests for KEMs on this branch:

https://github.com/prabhpreet/rosenpass/tree/liboqs-910

@aparcar I don't have permissions to write to your branch. Can you add the changes from the branch and see if the pipeline passes?

prabhpreet:
@aparcar Tests are passing with changes- please feel free to review #292 and incorporate the changes or review that PR and close this one

aparcar:
Superseded by #292

PR #298, opened 2024-05-01
wireguard-broker from dev/broker-architecture, remove old dependencies

prabhpreet:
WIP, step 1 to merging dev/broker-architecture


wireguard-broker/src/api/mio_client.rs
```diff
@@ -0,0 +1,205 @@
...
+    fn send_msg(&mut self, buf: &[u8]) -> Result<(), Self::SendError> {
+        self.flush()?;
+        self.send_or_buffer(&(buf.len() as u64).to_le_bytes())?;
+        self.send_or_buffer(&buf)?;
+        self.flush()?;
+
+        Ok(())
+    }
+
+    fn recv_msg(&mut self) -> Result<Option<&[u8]>, Self::RecvError> {
```
Rixxc, line 86: I think there is a special edge case where this function might not do what is expected of it.

For example:
1. call: reads the size of the message x. This resizes recv_buf to x.
2. call: tries to read in the message, but the message may actually be smaller than the size we received in the first call. So recv_off < recv_buf.len() and the function returns Ok(None).
3rd call: recv_off < recv_buf.len() is still true and therefore the stale message is not detected.

I suggest not to call the function recursively and explicitly reinstatiating the buffer after receiving a message.

Rixxc (changes requested):
Hi, thank you for this pr. The code is really well written.

I might have found an edge case where the `recv_msg` function might misbehave. Can you confirm that?

Otherwise the code looks good to me and can be merged.

prabhpreet:
Thanks  @Rixxc  and @koraa for the review. I added a feature flag, an integration test, removed PhantomData from BrokerServer and BrokerClient types (expected to have ownership of their inner types).

For @Rixxc 's comment on special edge cases:

> I think there is a special edge case where this function might not do what is expected of it.
> 
> For example:
> 
> 1. call: reads the size of the message x. This resizes recv_buf to x.
> 2. call: tries to read in the message, but the message may actually be smaller than the size we received in the first call. So recv_off < recv_buf.len() and the function returns Ok(None).
>    3rd call: recv_off < recv_buf.len() is still true and therefore the stale message is not detected.
> 
> I suggest not to call the function recursively and explicitly reinstatiating the buffer after receiving a message.

I introduced both current and expected states of buffer/length, removed recursion, fixed a constant buffer size instead of using vec for rx, and reset the state to expect length of buffer next on receiving shorter than expected buffer size. 

This makes me think of several followup tasks:

1. A more robust state machine with some sync bytes and state machine, and fuzz tests. Both client and server should accomodate the usecase of sockets over ssh, and failures that might arise from that. Maybe we should use something more standardized to send values like Postcard or bincode crates
2.  Storing anticipated PSKs in Secret<>, and removing use of Vec

wireguard-broker/src/api/mio_client.rs
prabhpreet: Thanks for this point. I made some changes and have a comment in the main thread to be able to track the conversation in case of rebasing the branch

koraa (approved):
Great work, thanks!

wireguard-broker/src/api/mio_client.rs
```diff
@@ -0,0 +1,231 @@
...
+                                "Oversized buffer ({len}) in psk buffer response."
+                            );
+
+                            self.recv_state = RxState::RxBuffer(0);
+                            self.expected_state = RxState::RxBuffer(len);
+                            continue;
+                        }
+                        RxState::RxBuffer(s) => {
+                            self.recv_state = RxState::RxSize(0);
+                            self.recv_state = RxState::RxSize(LEN_SIZE);
```
Rixxc, line 120: Do you intent to assign `recv_state` twice? I suspect the second assignment is meant for `expected_state`.

Rixxc (commented):
There is still one minor issue, that you should take a quick look at. Otherwise the pr looks good.

wireguard-broker/src/api/mio_client.rs
prabhpreet: Oops, didn't intend that. Thanks, good catch!

Rixxc (approved):

PR #303, opened 2024-05-07
feat(rosenpass): Add wireguard-broker interface in AppServer

prabhpreet:
Dynamically dispatch WireguardBrokerMio trait in AppServer. Also allows for mio event registration and poll processing, similar to dev/broker-architecture branch

Tested basic setup of connections between two devices


prabhpreet:
Reviewed with Karo on Matrix

Issue #245, opened 2024-02-28
bug: `rp explain` will throw error if no parameter is supplied to the command instead of explaining it


ilka-schulz:
I understand that `rp explain <subcommand>` will explain the specified subcommand. If I run this I do not get an explanation but the program tries to run the subcommand and complains about the missing parameter:

```text
$ ./rp explain genkey
FATAL: Required positional argument: PRIVATE_KEYS_DIR
Usage: ./rp genkey PRIVATE_KEYS_DIR
```

(on branch `main`, commit 8c469af6b14c6503a418dde236f856ff59492411)

prabhpreet:
@koraa With rp migrated to rust, we show this for rp explain genkey:
"WARN: the explain argument is no longer supported"
Required positional argument: PRIVATE_KEYS_DIR
Usage: rp genkey PRIVATE_KEYS_DIR

If we already have manpages, should we just remove this subcommand?

emilengler:
I think we can close this. Due to the removal of the feature, this plays no relevance anymore.

PR #313, opened 2024-05-16
chore(rp): Remove unsupported explain argument

prabhpreet:
Wrap up #245


emilengler:
I think we should leave explain with this notice in for the next few releases.

prabhpreet:
Closing, but will revert back to this in the coming releases

PR #321, opened 2024-05-23
Secret memory with memfd_secret

prabhpreet:
Implements:
- An additional allocator to use [memfd_secret(2)](https://man.archlinux.org/man/memfd_secret.2.en) and guard pages using [mmap(2)](https://man.archlinux.org/man/mmap.2), implemented in quininer/memsec#16
- An allocator that abstracts away underlying allocators, and uses specified allocator set by `rosenpass_secret_memory::policy` functions (or a function that sets `rosenpass_secret_memory::alloc::ALLOC_INIT`
- Updates to tests- integration, fuzz, bench: some tests use `procspawn` to spawn multiple processes with different allocator policies

PR #326, opened 2024-06-03
Make Rosenpass more verbose for a better user experience 

aparcar:
Add some extra logs here and there to make it easier to understand what's going on.

### DIFF
--- a/.ci/run-regression.sh
+++ b/.ci/run-regression.sh
@@ -1,33 +1,33 @@
 #!/bin/bash
 
-iterations=$1
-sleep_time=$2
+iterations="$1"
+sleep_time="$2"
 
-PWD=$(pwd)
-EXEC=$PWD/target/release/rosenpass
-LOGS=$PWD/output/logs
+PWD="$(pwd)"
+EXEC="$PWD/target/release/rosenpass"
+LOGS="$PWD/output/logs"
 
-mkdir -p output/logs
+mkdir -p "$LOGS"
 
 run_command() {
   local file=$1
   local log_file="$2"
-  ($EXEC exchange-config $file 2>&1 | sed "s/^/[$2] /" | tee -a $log_file) &
+  ("$EXEC" exchange-config "$file" 2>&1 | tee -a "$log_file") &
   echo $!
 }
 
 pids=()
 
-(cd output/dut && run_command "configs/dut-$iterations.toml" "dut.log") & piddut=$!
-for (( x=0; x<$iterations; x++ )); do
-  (cd output/ate && run_command "configs/ate-$x.toml" "ate-$x.log") & pids+=($!)
+(cd output/dut && run_command "configs/dut-$iterations.toml" "$LOGS/dut.log")
+for (( x=0; x<iterations; x++ )); do
+  (cd output/ate && run_command "configs/ate-$x.toml" "$LOGS/ate-$x.log") & pids+=($!)
 done
 
-sleep $sleep_time
+sleep "$sleep_time"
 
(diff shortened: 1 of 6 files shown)


prabhpreet (approved):

aparcar:
@koraa since this changes some basic behavior, could you please have a Quick Look if you agree to set the default from "Warning" to "Info" for logging?

koraa:
@aparcar

I don't have a strong opinion on this, to be honest. For most users trying Rosenpass for the first time, this makes things a bit more welcoming; for most admins with log files, this might pollute log files, so using info by default might be overly chatty. What we might really want is a minimal responsivity-minimal mode with basically a single "I started up, and I am listening" info in addition to the messages from warning and error.

Can you elaborate a bit on which user you are envisioning here? Which classes of user benefit, for whom does this introduce drawbacks.

Some target groups from the top of my head:

- Admin (of an org, professional)
- Dev
- User
- Beginner/Student

Use cases:

- Responsivity (I want to see my stuff works)
- Debugging (Something is not working for me)
- Logging (For retrospective debugging)
- Education


An admin, dev, and user/student have different needs regarding responsivity and logging and debugging.

Our current logging categories do not quite match these use cases and maybe we should start to change that.

@aparcar Also please do ping me in element when I should attention personally :)

koraa:
@aparcar @prabhpreet Also check out how the #321 branch interacts with this. Note that it adds a few informational messages about things that only the most specialized users from the Dev category would understand.

Showing these by default would lead to much confusion in most user categories.

We do not have agreement on what the info log level means right now…

aparcar:
> What we might really want is a minimal responsivity-minimal mode with basically a single "I started up, and I am listening" info in addition to the messages from warning and error.

That's what brought me to create this PR. I'm doing some benchmarks which includes spinning up 10-1000 Rosenpass instances. This happens via a script and right now all those instances stay entirely silent. A startup message would be good, so either we print it to stdout without a timestamp and keeping the logs on warning, or we use info by default and "downgrade" a few chatty infos to "debug" that would confuse the user.

The chattiest thing I see is the 2 minute key rotation information, so I'd downgrade it to "debug". By default the user doesn't need to see that information. Maybe we could add an info log that "the first" connection to a peer was successful, i.e. meaning "you did a fine configuration". I'll think a bit more about it.

I'd like to have a welcoming logging which covers all those user groups. Devs may go straight to debug log level but having a good first start experience seem important to me.

koraa:
> > What we might really want is a minimal responsivity-minimal mode with basically a single "I started up, and I am listening" info in addition to the messages from warning and error.
> 
> That's what brought me to create this PR. I'm doing some benchmarks which includes spinning up 10-1000 Rosenpass instances. This happens via a script and right now all those instances stay entirely silent. A startup message would be good, so either we print it to stdout without a timestamp and keeping the logs on warning, or we use info by default and "downgrade" a few chatty infos to "debug" that would confuse the user.
> 
> The chattiest thing I see is the 2 minute key rotation information, so I'd downgrade it to "debug". By default the user doesn't need to see that information. Maybe we could add an info log that "the first" connection to a peer was successful, i.e. meaning "you did a fine configuration". I'll think a bit more about it.
> 
> I'd like to have a welcoming logging which covers all those user groups. Devs may go straight to debug log level but having a good first start experience seem important to me.

Can we introduce new log levels? How about a "minimal-responsive" log level for now?

aparcar:
@koraa please give this another look.

koraa (approved):
Looks good to me, tests are failing though :)

aparcar:
Typo, sorry - fixed.

ilka-schulz:
closed due to inactivity

PR #336, opened 2024-06-07
chore: Rollback symbolic models to original state

koraa:
The later edits where unfortunately incomplete. They lacked modeling of multi-session, multi-user settings and they generally rendered the models less trustworthy from my perspective.

These edits are still interesting as a starting point for analyzing identity hiding and stealth, but they are not high-quality enough to be present in main.


pqcfox (approved):
Looks good to me--agree that these would be good to roll back, at least for the time being.

PR #338, opened 2024-06-10
drop unused import of WG_B64_LEN

aparcar:
This causes warnings

### DIFF
--- a/rp/src/exchange.rs
+++ b/rp/src/exchange.rs
@@ -2,7 +2,9 @@ use std::{net::SocketAddr, path::PathBuf};
 
 use anyhow::Result;
 
+#[cfg(any(target_os = "linux", target_os = "freebsd"))]
 use crate::key::WG_B64_LEN;
+
 #[derive(Default)]
 pub struct ExchangePeer {
     pub public_keys_dir: PathBuf,


prabhpreet:
the tests are failing

koraa:
@aparcar Is this ready?

aparcar:
It yet sorry

koraa:
Just checking because it was green! Thank you for the work you are doing here!

aparcar:
Sorry the issue was that some part are only used on `linux` and `freebsd`, but not macOS aka `darwin`.

I added a `cfg` option

rp/src/exchange.rs
```diff
@@ -2,7 +2,9 @@ use std::{net::SocketAddr, path::PathBuf};
 
 use anyhow::Result;
 
+#[cfg(any(target_os = "linux", target_os = "freebsd"))]
```
emilengler, line 5: I would suggest to add a warning suppression here instead of a `cfg` option, as constants do not influence the compile result. Besides, maintaining this seems to be more cumbersome than an `#[allow(unused)]`.

emilengler (changes requested):

koraa (approved):

rp/src/exchange.rs
koraa: Lets do that in another PR

PR #342, opened 2024-06-11
misc: add generate_configs.py script

aparcar:
The script can be used to simulate setups of different sizes. A short description is added to the `misc/` folder for further information.

This can be used for both benchmarking but also hunting down bugs which may occur with larger setups.


prabhpreet:
@aparcar Thanks for the scripts! It looks like the CI demands formatting to pass prettier. Could you please fix- we can merge them in then

aparcar:
I applied the prettier and will merge this once aarch64 is fully build.

prabhpreet (approved):

Issue #341, opened 2024-06-11
memfd-based allocator exhausts on more than 20 peers


prabhpreet:
I'm trying to run this commit with 50 peers and see the following error, could it be related?

```
[2024-06-10T15:18:50Z ERROR rosenpass_secret_memory::alloc::memsec::memfdsec] Allocation Layout { size: 524160, align: 1 (1 << 0) } was requested but memfd-based memsec returned a null pointer
memory allocation of 524160 bytes failed
```

_Originally posted by @aparcar in https://github.com/rosenpass/rosenpass/issues/321#issuecomment-2158643227_

prabhpreet:
Note to reproducer/fixer: 

Memory that can be allocated is limited by rlimit parameters when using memfd_secret- a soft limit by the process, and a hard limit by a privileged process. 

https://man7.org/linux/man-pages/man2/getrlimit.2.html

koraa:
> Note to reproducer/fixer:
> 
> Memory that can be allocated is limited by rlimit parameters when using memfd_secret- a soft limit by the process, and a hard limit by a privileged process.
> 
> https://man7.org/linux/man-pages/man2/getrlimit.2.html

Did we integrate the memfd-secret feature as an optional feature or is it enabled by default?

Why do different limits apply for memfd secret here? Is it due to the maximum allowed number of file descriptors?

koraa:
> > Did we integrate the memfd-secret feature as an optional feature or is it enabled by default?
> 
> Currently it is enabled by default. I'll put it behind a feature flag for the time being till we figure out this issue.
> 
> I did not anticipate this limit to be so low- we should definitely have a peer exhaustion test- there is likely an upper limit for malloc based allocation too.
> 
> > Why do different limits apply for memfd secret here? Is it due to the maximum allowed number of file descriptors?
> 
> From my understanding, the limit that applies is `RLIMIT_MEMLOCK`. In the conventional malloc strategy, the guard pages are not under `mlock` protection, rather are guarded by `mprotect`. Unfortunately, any space we allocate under memfd_secret is protected with `mlock` here including guard pages, so we hit the memlock limit quicker.
> 
> > ```
> >           This is the maximum number of bytes of memory that may be
> >           locked into RAM.  This limit is in effect rounded down to
> >           the nearest multiple of the system page size.  This limit
[...]

PR #343, opened 2024-06-11
Feature flag for memfd_secret alloc

prabhpreet:
Hide memfd_secret alloc behind feature flag, till we figure out a way to correct #341


pqcfox (approved):
Looks good to me, think this is a great stop-gap measure for now.

PR #349, opened 2024-06-14
feat: Regression CI based on misc/generate_configs.py

prabhpreet:


aparcar:
Very nice idea, I fixed some details and continued the work over at #363

Issue #215, opened 2023-12-15
Rp script should run as SUDO_USER


AliceOrunitia:
When `rosenpass-user` is not give, the parameter should default to the `SUDO_USER` environment variable.

pqcfox:
Would it be okay if I worked on this one @wucke13? Thanks!

Issue #214, opened 2023-12-15
`rp` script should not run as root


AliceOrunitia:
The `rp` script should accept `rosenpass-user` and `broker-user` parameters. `broker-user` defaults to `rosenpass-user`.
The script should use `capsh` ([see arch wiki](https://wiki.archlinux.org/title/capabilities#Running_a_program_with_temporary_capabilities))
to enable the broker to start with enough privileges to configure WireGuard.

pqcfox:
Would it be okay if I worked on this and #215 @wucke13? Thanks so much!

wucke13:
No objections from my side :)

pqcfox:
Two quick questions to make sure I understand the issue:

- For `rosenpass-user`, `rp` as written directly uses `AppServer` to set everything up. The best solution I can think of to incorporate `rosenpass-user` is to replace this and spawn a new process running `sudo -u <rosenpass-user> rosenpass exchange`--is there a cleaner way I'm missing, or is this an okay route?
- Similarly for `broker-user`, is it okay to just add `sudo -u <broker-user>` before the `wg` incantation in `NativeUnixBroker`? 

Thanks!

PR #351, opened 2024-06-19
Add `rosenpass-user` and `broker-user` parameters to `rp`

pqcfox:
This PR implements the requested features in #214 and #215, adding optional `rosenpass-user` and `broker-user` parameters to `rp`.

From #214:

> The rp script should accept rosenpass-user and broker-user parameters. broker-user defaults to rosenpass-user.
> The script should use capsh ([see arch wiki](https://wiki.archlinux.org/title/capabilities#Running_a_program_with_temporary_capabilities))
> to enable the broker to start with enough privileges to configure WireGuard.

From #215:

> When rosenpass-user is not given, the parameter should default to the SUDO_USER environment variable.

### DIFF
--- a/rp/src/cli.rs
+++ b/rp/src/cli.rs
@@ -1,5 +1,4 @@
-use std::path::PathBuf;
-use std::{iter::Peekable, net::SocketAddr};
+use std::{env, iter::Peekable, net::SocketAddr, path::PathBuf};
 
 use crate::exchange::{ExchangeOptions, ExchangePeer};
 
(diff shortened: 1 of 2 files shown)


koraa:
@pqcfox Can you split this into two commits? One for prettyfication one one for actual code changes? Makes it easier to review

koraa:
@pqcfox How do you verify that the indicated users are actually…used?

emilengler:
Environment variables should be documented. You can have a look at [this](http://man.openbsd.org/mdoc) page to see where the `ENVIRONMENT` section belongs and which macros to use. In case you have never written a manual page before or need help with this, please let me know! Their syntax can be very intimidating at first. 🙂

ilka-schulz:
closed due to inactivity

Issue #249, opened 2024-02-29
config: make `wg` field only available on binary builds, not on library builds


ilka-schulz:
see `rosenpass/src/config.rs`

PR #352, opened 2024-06-21
Include `RosenpassPeer`'s `wg` field only in binary builds 

pqcfox:
From #249:

> Title: make wg field only available on binary builds, not on library builds
> see `rosenpass/src/config.rs

Since conditional compilation for crate type isn't supported yet in Rust, the next best options are:

- Create a secondary `Cargo.toml` in a new folder, e.g. `rosenpass/rlib`, and point everything depending on the `rosenpass` library to that directory instead (see [this blog post](https://osa1.net/posts/2020-12-24-crate-type-conditional.html)), using a `default` feature to control conditional compilation. 
- Keep the structure as is, and add a feature into `required-features` under `[[bin]]` in the existing Cargo.toml.
- Seperate the `bin` and `lib` parts of `rosenpass` into separate crates, and use `default` features in the `bin` crate to distinguish the two.

The first option is simplest, but is somewhat hacky and unintuitive for someone trying to build the library. The second option works if someone only builds the binary or library half of `rosenpass` at once, but requires them to manually add a feature in the former case, and this option entirely fails if someone builds both at once (the library ends up with the `wg` field).

As such, I went with the third option--would love feedback on this choice, especially as it relates to any possible issues with distributing the Rosenpass crate.


ilka-schulz:
closed due to inactivity

PR #363, opened 2024-07-08
Regression CI and fixup

aparcar:


aparcar:
This now also verifies that 100 keys were indeed exchanged.

aparcar:
@prabhpreet thanks again for your initial work on this!

.ci/run-regression.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
emilengler, line 1: `#!/usr/bin/env bash` please for better compatibility with Nix.

.ci/run-regression.sh
```diff
@@ -0,0 +1,33 @@
...
+
+iterations="$1"
+sleep_time="$2"
+
+PWD="$(pwd)"
+EXEC="$PWD/target/release/rosenpass"
+LOGS="$PWD/output/logs"
```
emilengler, line 8: Maybe `mktemp(1)` is better than an `output/` directory? Just a suggestion, I do not have a strong opinion about either approach.

misc/generate_configs.py
```diff
@@ -1,14 +1,14 @@
...
-
+import os
 
 config = dict(
     peer_counts=[1, 5, 10, 50, 100, 500],
     peer_count_max=100,
-    ate_ip="192.168.2.1",
-    dut_ip="192.168.2.4",
+    ate_ip="127.0.0.1",
+    dut_ip="127.0.0.1",
```
emilengler, line 9: Can we test IPv6 as well?

emilengler (changes requested):

koraa (approved):

misc/generate_configs.py
aparcar: Sure but it wasn't part of the regression

.ci/run-regression.sh
aparcar: Since it's running in the CI, it's a fresh folder anyway.

.ci/run-regression.sh
aparcar: Fixed.

aparcar:
@emilengler I think this is ready, please feel free to merge

emilengler:
could you update the branch?

PR #376, opened 2024-07-25
Add broker support to Rosenpass using `MioBrokerClient` (backport of dev/broker-architecture)

pqcfox:
This PR adds a `enable_wg_broker`  feature to the `rosenpass` binary, which enables two new flags: `--psk_broker_path` for providing a path to a socket for a WireGuard broker, and `--psk_broker_fd` for providing a file descriptor instead (useful for sharing a socket among multiple processes).


pqcfox:
(Note that cargo failures above are due to `prettier` check fails.)

pqcfox:
Converted back to draft, since the branch as is is panicking on main. Preliminary review is still very welcome though!

pqcfox:
Figured out what was causing the panic and fixed it, but now when I try to exchange configurations between two Rosenpass instances, it hangs when one has a broker. Continuing to try to fix that.

pqcfox:
One note to add to the above--currently, `--psk-broker-spawn` is used to set up a socket pair and spawn a broker in a new thread. The reason this isn't default behavior (like it is in `dev/broker-architecture`) is that I still need how to put `rosenpass-wireguard-broker-socket-handler` in the path when running `cargo test --all-features` from CI, which I think is worth giving a bit of thought and placing in a separate PR.

rosenpass/src/cli.rs
```diff
@@ -365,6 +437,90 @@ impl CliCommand {
...
+        Ok(Box::new(NativeUnixBroker::new()))
+    }
+
+    #[cfg(feature = "enable_wg_broker")]
+    fn get_broker_socket(broker_interface: BrokerInterface) -> Result<UnixStream, anyhow::Error> {
+        // Connect to the psk broker unix socket if one was specified
+        // OR OTHERWISE spawn the psk broker and use socketpair(2) to connect with them
+        match broker_interface {
+            BrokerInterface::Socket(broker_path) => {
+                let sock = net::UnixStream::connect(broker_path)?;
```
koraa, line 472: Why not use `mio::net::UnixStream::connect` here?

rosenpass/src/cli.rs
```diff
@@ -365,6 +437,90 @@ impl CliCommand {
...
+                sock.set_nonblocking(true)?;
+                Ok(UnixStream::from_std(sock))
+            }
+            BrokerInterface::FileDescriptor(broker_fd) => {
+                // mio::net::UnixStream doesn't implement From<OwnedFd>, so we have to go through std
+                let sock = net::UnixStream::from(claim_fd(broker_fd)?);
```
koraa, line 478: Can you also set the close on exec flag here? (You will probably need to use the fcntl support from rustix).

wireguard-broker/src/brokers/mio_client.rs
```diff
@@ -158,10 +160,17 @@ impl BrokerClientIo for MioBrokerClientIo {
                 {
                     let bytes = raw_recv(&self.socket, &mut self.recv_buf[x..y])?;
 
+                    // If we've received nothing so far, and raw_recv came up empty,
```
koraa, line 163: Why was this neccesary?

koraa (changes requested):
Looks good generally! A couple of comments…

Could you also write a quick guide about how to manually test the features of this PR before merge?

koraa (changes requested):
Could you also please make sure that there is a separate CI run with the broker features enabled?

koraa (changes requested):
Could you also please delete the path based listen sockets at program shutdown (through a Drop implementation)

koraa:
> Could you also please delete the path based listen sockets at program shutdown (through a Drop implementation)

Actually, strike that…it seems to be exceedingly difficult.

koraa:
> Could you also please make sure that there is a separate CI run with the broker features enabled?

From what it looks like, we run with --all-features, so it should be enabled…

wireguard-broker/src/brokers/mio_client.rs
pqcfox: I noticed that Rosenpass would generally panic as soon as it started to exchange with a peer; the first Mio poll that happened would always report that nothing had been received yet, resulting in a state of `RxState::RxSize(0)` and causing the `bail!("Invalid state")` below to trigger. This would bubble up through to the top of the program, causing a panic.

pqcfox:
Perfect, made sure that the above is working manually! I've also included a doc for testing in `doc/broker/broker_testing.md`.

PR #391, opened 2024-08-07
add test cases for util modules

aparcar:


blipp:
Idea for going forward: use an `mktemp` function for temporary test files

aparcar:
The nix failure seems unrelated to me

aparcar:
Any ideas why codecov isn't run here?

blipp:
Maybe because it is a draft PR now?

PR #426, opened 2024-09-17
docs: Correct protocol retransmission unit/vars

aparcar:
Those are seconds not ms, also it's BEGIN not BEG.

### DIFF
--- a/rosenpass/src/protocol/protocol.rs
+++ b/rosenpass/src/protocol/protocol.rs
@@ -134,11 +134,10 @@ pub const PEER_COOKIE_VALUE_EPOCH: Timing = 120.0;
 // decryption for a second epoch
 pub const BISCUIT_EPOCH: Timing = 300.0;
 
-// Retransmission pub constants; will retransmit for up to _ABORT ms; starting with a delay of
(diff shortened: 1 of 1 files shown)


aparcar:
@koraa RETRANSMIT_ABORT is actually nowhere used in the code. Should this be fixed or is the variable obsolete?

aparcar:
I just added a patch to print the retransmission time and it works fine, I wonder if we should set the retransmission abort to a counter instead of a specific duration. We could just compare it with `ih.tx_count`?

aparcar:
Also I'm wondering if the comment regarding the use of `rand` crate is still valid. The same seems to be used in `public.rs` without the comment.

koraa (approved):

koraa:
@aparcar Which comment about the rand crate? Lets remove the retransmission abort value if its not being used…

aparcar:
> @aparcar Which comment about the rand crate? 

https://github.com/rosenpass/rosenpass/blob/main/rosenpass/src/protocol/protocol.rs#L1476

> Lets remove the retransmission abort value if its not being used…

I'll add another commit

aparcar:
I added all requested changes, please feel free to merge at your convenience

PR #430, opened 2024-09-25
wip: Rewrite to more modular CLI

aparcar:
This is a rewrite of the CLI to make it more modular and easier to maintain. Initial motivation was to create `man` pages automatically. 

The work started prior the *broker* API stuff, so that part is missing right now - if the general structure is acceptable I'd implement that, too.

Some assumptions:
* Use `clap_mangen` to generate documentation instead of "raw" files.
* Use `clap-verbosity-flag` for better handling log levels. I.e. you add `-q` for quite or `-vv` to be more verbose than `-v`.
* Have `cli.rs` as abstract thing only, don't add any functionality to it.
* Move `event_loop.rs` into its own file (included via `cmd/exchange.rs` and `cmd/exchangeconfig.rs`). Should it be added to `cmd/` instead?
* ~~Dropped deprecated `Keygen` command~~

**I expect pretty much all tests to fail at this stage**

I think with the [derive] approach it's not fully possible to add sections like "BUGS" and "STANDARDS", I may use partly the "builder" approach or create upstream PRs against `clap`.


koraa:
Thank you for submitting this PR; it looks very nice.

What worries me a bit is the lack of integration tests checking whether the CLI keeps functioning, so its hard to figure out whether this change could introduce any regression.

> Dropped deprecated Keygen command

Lets keep it for now for backwards compatibility. There is quite a number of deployments at this point so its quite important not to throw away stuff unless its really necessary.

aparcar:
@koraa this seems to slowly getting into shape, one thing I'm currently troubled with is the dynamic import of the ApiCli. Using the current approach for `build.rs`, this causes errors. I'm not familiar enough with the build system here so if anyone has an idea, please help out.

aparcar:
As discussed via chat I dropped `build.rs` and moved it to `main.rs`, which now allows to use `--print-manpage` and `--print-completions`, which was a long hanging fruit anyway.

By using `arg_required_else_help` from `clap`, the help page is printed by default which was a outstanding TODO anyway https://github.com/rosenpass/rosenpass/blob/main/rosenpass/src/cli.rs#L223

I took the freedom to call the `get_brooker_interface` only when actually used, i.e. when `exchange` or `exchange-config` is called.

aparcar:
For packaging, this gives a good idea how OS could handle completions and man pages: https://gitlab.archlinux.org/archlinux/packaging/packages/genact/-/blob/main/PKGBUILD?ref_type=heads

ilka-schulz:
closed due to inactivity

PR #431, opened 2024-09-25
Format Cargo files

aparcar:


aparcar:
Cargo audit doesn't like 0.0.0 since there were security issues in the past with versions like 0.2.0. Another approach would be to use a version like 999.9.9?

cvengler:
There is also an (internal?) wiki page describing the release workflow, although I cannot find it right now. I would recommend mentioning to change the tag to the release and to 0.0.0 afterwards.

But instead of using `0.0.0` why not add a `-dev` prefix?

aparcar:
So you'd go for 0.2.2-dev?

aparcar:
I dropped the version commit so we weg the formatting in. Could this please be approved?

koraa (approved):

koraa (approved):

PR #432, opened 2024-10-01
ci: add `concurrency` option to skip in progress

aparcar:
Instead of running outdated CI jobs, skip them automatically.


aparcar:
This should be merged to massively lower CI delays :)

PR #437, opened 2024-10-07
RFC feat(rosenpass): git describe based version

aparcar:
Use `git describe` to automatically generate the version, instead of using "just" cargo. The advantage is to have a dynamic version showing the exact git hash and commits since latest release.

### DIFF
--- a/rosenpass/build.rs
+++ b/rosenpass/build.rs
@@ -47,6 +47,11 @@ fn man() {
 fn main() {
     // For now, rerun the build script on every time, as the build script
     // is not very expensive right now.
+
+    let output = Command::new("git").args(&["describe"]).output().unwrap();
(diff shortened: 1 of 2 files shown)


wucke13:
This creates a conflict with our Nix setup. From within the Nix build, the git repo context (so everything in `.git`) is intentionally not available. This is done, because `.git` repos are non-deterministic in their representation, hence one would put reproducibility in jeopardy when exposing the `.git` dir to builders. So, it would be best if this feature fails gracefully upon not finding a `.git` dir. I can also patch out the issue in the Nix expressions, but generally I find it unpolite of build processes to presume that they are executed within a `git` repo.

aparcar:
Good call, thanks for the clarification.

PR #453, opened 2024-10-21
ci: add regression test for boot race condition

aparcar:
If two instances start up at the same time, they end up with different keys on both ends. Test this with different delays of 2 (working), 1 (flaky) and 0 (broken) seconds.

### DIFF
--- a/.ci/boot_race/a.toml
+++ b/.ci/boot_race/a.toml
@@ -1,0 +1,14 @@
+public_key = "rp-a-public-key"
+secret_key = "rp-a-secret-key"
+listen = ["127.0.0.1:9999"]
+verbosity = "Verbose"
+
+[api]
+listen_path = []
+listen_fd = []
+stream_fd = []
+
+[[peers]]
+public_key = "rp-b-public-key"
+endpoint = "127.0.0.1:9998"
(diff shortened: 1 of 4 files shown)


aparcar:
Since this is a rare corner case it'll be handled with low priority.

An initial discussion about this bug suggested to compare the result of a final `InitConf` hash and if both server and client are in the middle of a handshake, take the hash that's *larger* (i.e. higher integer value). This would drop the *smaller* handshake if happening in parallel without giving either communication partner the ability to force their handshake to be preferred.

koraa:
Can you check if there is some way to mark this as failure-expected so we can merge this for now?

koraa:
(Rebased)

aparcar:
> Can you check if there is some way to mark this as failure-expected so we can merge this for now?

Doesn't look like https://github.com/actions/runner/issues/2347#issuecomment-2301892819

I'd just leave it open for now?

aparcar:
I changed it so the job itself doesn't fail but instead adds a warning.

Once the race condition is fixed, we can set it to `exit 1`.

From my side this is now ready to fly.

image

koraa (approved):

PR #461, opened 2024-10-28
ci(dependabot): check for GitHub action updates

aparcar:
We already use Dependabot for cargo updates, use it for GitHub action updates, too. Right now we see warnings every now and then because Node wants another upgrade or some checkout stuff is about to be deprecated.


blipp (approved):
thanks!

PR #465, opened 2024-10-30
fix(dualstack): logic error in IPv4 socket creation

### LINKED ISSUE #448: send_scouting function: Iterator restart broken
Hi,

in function `send_scouting` in file https://github.com/rosenpass/rosenpass/blob/de51c1005f476d15f242fce9d9903245307b718f/rosenpass/src/app_server.rs#L493, it can happen that the iterator of the inner loop appears empty if a foregoing iteration of the outer loop resulted in an `Err(...)` result in the inner loop.

I triggered this by setting the listen address (eronneously) to `[::1]:<port>` in my rosenpass config.
(Everything works with `[::]:<port>` as listen address, but this was still considered a bug by @koraa in our conversation today)

It appears that the "iterator restart" is broken, as `collect`ing into a new `Vec` fixes the problem.

aparcar:
This commit fixes a logic error in the socket creation for v6 and v4. The error case is that an IPv6 only socket would prevent an additional IPv4 socket to be created.

With this commit each variable it's made sure that at least a dualstack IPv6 socket exists or an IPv4 and IPv6 socket.

Special case is macOS, while they report to support dualsocket, in fact it doesn't work.

Move the logic to it's own function and add some test cases.

Fixes: https://github.com/rosenpass/rosenpass/issues/448

### DIFF
--- a/rosenpass/src/app_server.rs
+++ b/rosenpass/src/app_server.rs
@@ -70,6 +70,51 @@ fn ipv6_any_binding() -> SocketAddr {
     SocketAddr::V6(SocketAddrV6::new(IPV6_ANY_ADDR, 0, 0, 0))
 }
 
+fn open_sockets(sockets: &mut Vec<mio::net::UdpSocket>) -> anyhow::Result<()> {
+    macro_rules! try_register_socket {
+        ($title:expr, $binding:expr) => {{
+            let r = mio::net::UdpSocket::bind($binding);
+            match r {
+                Ok(sock) => {
+                    sockets.push(sock);
+                    Some(sockets.len() - 1)
+                }
+                Err(e) => {
+                    warn!("Could not bind to {} socket: {}", $title, e);
+                    None
+                }
+            }
+        }};
+    }
+
+    let mut have_ipv4 = false;
+    let mut have_ipv6 = false;
+
+    sockets.iter().for_each(|s| {
+        let addr = s.local_addr().unwrap();
+        if addr.is_ipv4() {
+            have_ipv4 = true;
+        } else {
+            have_ipv6 = true;
+            if !s.only_v6().unwrap_or(true) && std::env::consts::OS != "macos" {
+                have_ipv4 = true;
+            }
+        }
+    });
+
+    if !have_ipv4 {
+        log::debug!("Register IPv4 socket as fallback.");
+        try_register_socket!("IPv4", ipv4_any_binding());
+    }
+
+    if !have_ipv6 {
+        log::debug!("Register IPv6 socket as fallback.");
+        try_register_socket!("IPv6", ipv6_any_binding());
+    }
+
+    Ok(())
+}
+
 #[derive(Debug, Default)]
 pub struct MioTokenDispenser {
     counter: usize,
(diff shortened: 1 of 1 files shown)


KarlK90 (commented):
Hi @aparcar,

This change alters the default behavior of rosenpass IMHO unexpected ways e.g.: If I want to only listen on the IPv6 loopback interface on port 9998 by setting `listen = ["[::1]:9998"]` I now get a IPv4 socket which listens on all interfaces as well, possibly exposing it on interfaces I don't want it exposed.

After giving it some thought I do believe that the previous behavior is in fact correct - just that rosenpass should provide more hints how it is reachable on startup - possibly with a table which indicates which sockets are ipv4/ipv6/dual-stack so it is more obvious why a connection attempt might not succeed.

rosenpass/src/app_server.rs
```diff
@@ -70,6 +70,51 @@ fn ipv6_any_binding() -> SocketAddr {
...
+        let addr = s.local_addr().unwrap();
+        if addr.is_ipv4() {
+            have_ipv4 = true;
+        } else {
+            have_ipv6 = true;
+            if !s.only_v6().unwrap_or(true) && std::env::consts::OS != "macos" {
+                have_ipv4 = true;
+            }
+        }
+    });
```
KarlK90, line 103: This block is almost verbatim repeated in `helper_support_ipv4_ipv6` it could be a helper function which is shared between tests and actual code.

rosenpass/src/app_server.rs
```diff
@@ -70,6 +70,51 @@ fn ipv6_any_binding() -> SocketAddr {
...
+
+    if !have_ipv4 {
+        log::debug!("Register IPv4 socket as fallback.");
```
KarlK90, line 106: If the change in behavior is accepted in rosenpass - I would promote it to `info!` and give it a descriptive text explaining why the ipv4 socket was registered.

rosenpass/src/app_server.rs
```diff
@@ -70,6 +70,51 @@ fn ipv6_any_binding() -> SocketAddr {
...
+                }
+            }
+        }};
+    }
+
+    let mut have_ipv4 = false;
+    let mut have_ipv6 = false;
+
+    sockets.iter().for_each(|s| {
+        let addr = s.local_addr().unwrap();
```
KarlK90, line 94: Please replace the `unwrap()` call by skipping this socket.

rosenpass/src/app_server.rs
```diff
@@ -70,6 +70,51 @@ fn ipv6_any_binding() -> SocketAddr {
...
+        }
+    });
+
+    if !have_ipv4 {
+        log::debug!("Register IPv4 socket as fallback.");
+        try_register_socket!("IPv4", ipv4_any_binding());
+    }
+
+    if !have_ipv6 {
+        log::debug!("Register IPv6 socket as fallback.");
```
KarlK90, line 111: Same here.

rosenpass/src/app_server.rs
```diff
@@ -1288,3 +1303,62 @@ impl crate::api::mio::MioManagerContext for MioManagerFocus<'_> {
...
+
+    #[test]
+    pub fn test_open_sockets_ipv4_localhost() {
+        let mut sockets = Vec::new();
+        sockets.push(
+            mio::net::UdpSocket::bind(SocketAddr::V4(SocketAddrV4::new(
+                Ipv4Addr::new(127, 0, 0, 1),
+                0,
+            )))
+            .unwrap(),
```
KarlK90, line 1342:
```suggestion
        sockets.push(UdpSocket::bind("127.0.0.1:0".parse().unwrap()).unwrap());
```

Nit, IMHO this is more readable. Same for the IPv6 case.

rosenpass/src/app_server.rs
```diff
@@ -1288,3 +1303,62 @@ impl crate::api::mio::MioManagerContext for MioManagerFocus<'_> {
...
+                }
+            }
+        }
+        supports_iv4 && supports_ipv6
+    }
+
+    #[test]
+    pub fn test_open_sockets_empty() {
+        let mut sockets = Vec::new();
+        let _ = open_sockets(&mut sockets);
```
KarlK90, line 1330:
```suggestion
        assert!(open_sockets(&mut sockets).is_ok());
```

You probably want to test that your function didn't error out early.

KarlK90 (changes requested):

aparcar:
First of, there seem to be a logic flaw in the current implementation. The need for a v4 socket is only evaluated if no `listen` address is defined at all. The functionality is "hidden" in the `is_empty()` block, see [here](https://github.com/rosenpass/rosenpass/blob/main/rosenpass/src/app_server.rs#L601). This renders Rosenpass broken for configuration like described by @tfc. 

Looking at the WireGuard [Kernel](https://github.com/torvalds/linux/blob/master/drivers/net/wireguard/socket.c#L20) and [Rust](https://github.com/cloudflare/boringtun/blob/64a2fc7c63e5f5a9aa5b8e1fbcefd504778ad118/boringtun/src/device/peer.rs#L123) implementation, both listen on any IPv4 and IPv6 port by default. Since, just like WireGuard, Rosenpass must not respond to anything but valid messages, containing message data encrypted with the receivers public key, it looks like the same "expose" level to me.

To my understanding we're facing the following problem which isn't even super unlikely: A user defines a remote server as peer available over IPv6 and IPv4. Later the user adds an IPv4 listen address and thereby looses IPv6 connectivity to remote peers. This behavior seems rather confusing and therefore I suggest to decouple both logics entirely. After all, it's an outgoing socket to reach remote peers, Mio shouldn't even accept initiation messages there.

ilka-schulz:
closed due to inactivity

PR #490, opened 2024-11-11
refactor: replace `rustix` with `std` where possible

mkroening:
This reduces the amount of code that relies on `rustix`.
The FD traits can just be imported from `std` instead of `rustix`.
Closing FDs can also be done via `std`.
After gating `rustix` in `rosenpass` on `experiment-api`, the only non-optional use of `rustix` is in `util::fd`.


koraa:
Thank you!

Issue #487, opened 2024-11-08
RFC: How to open connecting (outgoing) sockets


aparcar:
Hi, right now Rosenpass supports being a client, a server or both. 

* Client if a peer has an `endpoint` defined, so Rosenpass tries to connect to it.
* Server if one or more `listen` addresses are given. In that case the sockets are opened.
* Client and server if at least one peer has an `endpoint` defined and at least one `listen` address is set.

The opening of connecting (outgoing aka to other peers) is influenced by the value of the `listen` option:

* If `listen` is empty, a `[::]` (dual stack) socket with random port is opened. It may connect to whatever IP address the device running Rosenpass can connect to. This opens at the same time the random port for incoming connections.
* If `listen` has at least one IP and port defined, listen on that specific interface and try to connect from that socket to whatever peer address.

The first case may result in unexpected open ports, even though the user did not specify `listen`, it still listens on all interfaces/ports. The second case may result in cases where a user listens only on IPv4 IP addresses but a peer's endpoint is IPv6 only. The local Rosenpass socket won't be able to connect, similar as reported in #448.

Looking at how Wireguard (Kernel + Rust) implementations handle this, both separate the receiving and the sending logic. If a listen port is defined, that port is opened on all interfaces. Outgoing connections use a random port. This approach comes with maximal usability from a user perspective and as WireGuard (just like Rosenpass) never answers on packages it doesn't recognize, it's "stealthy".

The current implementation tries to be smart and secure, but may result in really unexpected behavior. My suggestion would be to separate the incoming and outgoing logic for sockets.

PR #496, opened 2024-11-14
Introduce "High Quality" Systemd unit files for `rosenpass` and `rp`

tfc:
This PR introduces systemd unit files to simplify running `rosenpass` and`rp` with minimal setup. The workflow after these changes is like this:

1. Create configuration files, private/public keys on both server and client, and distribute them accordingly
2. Start the service using `systemctl start rp@<config-name>.service` (or respectively `rosenpass@<config-name>.service`)
3. Enjoy a post-quantum-secure encrypted connection

I tested this manually on an Ubuntu machine. The automated tests use NixOS, see next section.

## Changes Introduced

To achieve the above, the following changes have been made:

- Rosenpass Unit File: 
  - Added a systemd unit file for `rosenpass`
  - Implemented a NixOS integration test for verification

- `rp` App: Patched the `rp` application to support configuration files

- `rp` Unit File:
  - Added a systemd unit file for `rp`
  - Added a NixOS integration test for verification

## Considerations and Notes

While working on this, a few challenges and limitations were encountered:

- Rust Code Quality: The Rust code may not be best practice and could be improved. Feedback on how to refine it is appreciated.
- `rosenpass` Client Config Parsing: The `listen` field is currently required in the client configuration, which feels wrong. However, i feel addressing this would be feature-creep in this PR.
- `rp` IP Parameter Limitation: The `ip` parameter for `rp exchange` could theoretically support multiple IPs, but this functionality was not implemented in this iteration, as it already supports the standard scenarios.

The need to extend the `rp` code was not anticipated during the initial planning of this work package. However, the current implementation is functional for simple use cases and should provide a solid foundation for further enhancements if needed.

I hope this implementation aligns with your goals and expectations. Feedback and suggestions for improvement are welcome!

### DIFF
--- a/rp/src/cli.rs
+++ b/rp/src/cli.rs
@@ -12,6 +12,9 @@ pub enum Command {
         public_keys_dir: PathBuf,
     },
     Exchange(ExchangeOptions),
+    ExchangeConfig {
+        config_file: PathBuf,
+    },
     Help,
 }
 
@@ -19,6 +22,7 @@ enum CommandType {
     GenKey,
     PubKey,
     Exchange,
+    ExchangeConfig,
 }
 
 #[derive(Default)]
@@ -33,8 +37,9 @@ fn fatal<T>(note: &str, command: Option<CommandType>) -> Result<T, String> {
             CommandType::GenKey => Err(format!("{}\nUsage: rp genkey PRIVATE_KEYS_DIR", note)),
             CommandType::PubKey => Err(format!("{}\nUsage: rp pubkey PRIVATE_KEYS_DIR PUBLIC_KEYS_DIR", note)),
             CommandType::Exchange => Err(format!("{}\nUsage: rp exchange PRIVATE_KEYS_DIR [dev <device>] [ip <ip1>/<cidr1>] [listen <ip>:<port>] [peer PUBLIC_KEYS_DIR [endpoint <ip>:<port>] [persistent-keepalive <interval>] [allowed-ips <ip1>/<cidr1>[,<ip2>/<cidr2>]...]]...", note)),
+            CommandType::ExchangeConfig => Err(format!("{}\nUsage: rp exchange-config <CONFIG_FILE>", note)),
         },
-        None => Err(format!("{}\nUsage: rp [verbose] genkey|pubkey|exchange [ARGS]...", note)),
+        None => Err(format!("{}\nUsage: rp [verbose] genkey|pubkey|exchange|exchange-config [ARGS]...", note)),
     }
 }
 
@@ -252,6 +257,21 @@ impl Cli {
 
                     let options = ExchangeOptions::parse(&mut args)?;
                     cli.command = Some(Command::Exchange(options));
+                }
+                "exchange-config" => {
+                    if cli.command.is_some() {
+                        return fatal("Too many commands supplied", None);
+                    }
+
+                    if let Some(config_file) = args.next() {
+                        let config_file = PathBuf::from(config_file);
+                        cli.command = Some(Command::ExchangeConfig { config_file });
+                    } else {
+                        return fatal(
+                            "Required position argument: CONFIG_FILE",
+                            Some(CommandType::ExchangeConfig),
+                        );
+                    }
                 }
                 "help" => {
                     cli.command = Some(Command::Help);
--- a/rp/src/cli.rs
+++ b/rp/src/cli.rs
@@ -32,7 +32,7 @@ fn fatal<T>(note: &str, command: Option<CommandType>) -> Result<T, String> {
         Some(command) => match command {
             CommandType::GenKey => Err(format!("{}\nUsage: rp genkey PRIVATE_KEYS_DIR", note)),
             CommandType::PubKey => Err(format!("{}\nUsage: rp pubkey PRIVATE_KEYS_DIR PUBLIC_KEYS_DIR", note)),
-            CommandType::Exchange => Err(format!("{}\nUsage: rp exchange PRIVATE_KEYS_DIR [dev <device>] [listen <ip>:<port>] [peer PUBLIC_KEYS_DIR [endpoint <ip>:<port>] [persistent-keepalive <interval>] [allowed-ips <ip1>/<cidr1>[,<ip2>/<cidr2>]...]]...", note)),
+            CommandType::Exchange => Err(format!("{}\nUsage: rp exchange PRIVATE_KEYS_DIR [dev <device>] [ip <ip1>/<cidr1>] [listen <ip>:<port>] [peer PUBLIC_KEYS_DIR [endpoint <ip>:<port>] [persistent-keepalive <interval>] [allowed-ips <ip1>/<cidr1>[,<ip2>/<cidr2>]...]]...", note)),
         },
         None => Err(format!("{}\nUsage: rp [verbose] genkey|pubkey|exchange [ARGS]...", note)),
     }
@@ -142,6 +142,13 @@ impl ExchangeOptions {
                         options.dev = Some(device);
                     } else {
                         return fatal("dev option requires parameter", Some(CommandType::Exchange));
+                    }
+                }
+                "ip" => {
+                    if let Some(ip) = args.next() {
+                        options.ip = Some(ip);
+                    } else {
+                        return fatal("ip option requires parameter", Some(CommandType::Exchange));
                     }
                 }
                 "listen" => {
--- a/systemd/rp@.service
+++ b/systemd/rp@.service
@@ -1,0 +1,48 @@
+[Unit]
+Description=Rosenpass key exchange for %I
+Documentation=man:rosenpass(1)
+Documentation=https://rosenpass.eu/docs
+
+After=network-online.target nss-lookup.target
+Wants=network-online.target nss-lookup.target
+PartOf=rosenpass.target
+
+[Service]
+ExecStart=rp exchange-config /etc/rosenpass/%i.toml
+LoadCredential=pqpk:/etc/rosenpass/%i/pqpk
+LoadCredential=pqsk:/etc/rosenpass/%i/pqsk
+LoadCredential=wgsk:/etc/rosenpass/%i/wgsk
+
+AmbientCapabilities=CAP_NET_ADMIN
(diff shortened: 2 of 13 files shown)


koraa (commented):
Looks pretty good!

koraa:
Darwin (the system underlying mac) is failing; something about the futures crate.

We don't really support darwin/mac right now so I do not believe making sure everything works perfectly in darwin is worth the effort here. However, we do need to make the CI pass…

systemd/rp@.service
```diff
@@ -0,0 +1,48 @@
...
+Documentation=man:rosenpass(1)
+Documentation=https://rosenpass.eu/docs
+
+After=network-online.target nss-lookup.target
+Wants=network-online.target nss-lookup.target
+PartOf=rosenpass.target
+
+[Service]
+ExecStart=rp exchange-config /etc/rosenpass/%i.toml
+LoadCredential=pqpk:/etc/rosenpass/pqpk
```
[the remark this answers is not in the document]
tfc: I screwed up my rebase. My fault. It is all in order now.

tests/systemd/rp.nix
```diff
@@ -0,0 +1,128 @@
+{ pkgs, ... }:
```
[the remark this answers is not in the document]
tfc: Yes. The tests are both installed in the flake `checks` which means they are at least run by `nix flake check`

tfc:
> Darwin (the system underlying mac) is failing; something about the futures crate.
> 
> [..] However, we do need to make the CI pass…

Isn't that then a problem in all the other PRs, too?

tests/systemd/rp.nix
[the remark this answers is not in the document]
koraa: Thanks!

koraa:
> > Darwin (the system underlying mac) is failing; something about the futures crate.
> > [..] However, we do need to make the CI pass…
> 
> Isn't that then a problem in all the other PRs, too?

Well, the other PRs haven't triggered the problem recently.

koraa:
> > Well, the other PRs haven't triggered the problem recently.
> 
> Looking at the error messages, it seems like both rosenpass and rp trigger this. I wonder why only in this PR, as it only patches rp and not rosenpass.
> 
> I also looked at how to fix it and make it build on my macbook, which was successful, however looking at policy decisions like [this one](https://github.com/rosenpass/rosenpass/blob/main/rp/Cargo.toml#L27) without the knowledge which platforms _are_ supported, i am not sure if the `futures = ...` line should simply be pulled out of the conditional block or if there is a better solution.

Honestly, I have no idea. The `futures` crate is not used outside of the `rp` subcrate.

I do not see a compelling reason to not pull futures out of the block; the rp subcrate was not written to the same quality standard as the rest of rosenpass, so I am not too surprised about some noise.

> > We don't really support darwin/mac right now so I do not believe making sure everything works perfectly in darwin is worth the effort here. However, we do need to make the CI pass…
> 
> I wonder why is darwin then even activated in the CI? Is that by accident?

Someone contributed a feature and then abandoned it. We haven't had enough compelling reasons to throw it out again.

tfc:
@koraa Thanks for looking into the PR!

I’ve pushed a fix that makes it build on my M3 MacBook (aarch64). However, the following CI jobs are still failing:

- Nix / Build x86_64-darwin.rosenpass
- Nix / Build x86_64-darwin.rp
- QC / cargo-test (macos-13)

The first two are clearly for x86_64. For the third one, I went through the logs but couldn’t figure out for sure if it’s also x86_64. It seems likely, though.

Unfortunately, I don’t have an x86_64 Mac to debug this. From what I can see in the logs, it looks like the `futures` dependency isn’t being used on darwin-x86_64, even though I changed it to apply to all platforms.

Since macOS isn’t a main target for the project, and fixing this would take extra time and tools I don’t have right now, I feel like this is out of scope for my work here. I’m usually happy to help beyond what I’m directly responsible for, but this issue is starting to block the tasks I was actually hired to do.

If someone with the right setup can take a closer look, I’m happy to help however I can to move this forward. Let me know how we can proceed!

koraa:
@tfc Thank you for your work.

I took a look at the issue and the problem was luckily easy to fix by adding the `futures` crate to the list of dependencies unconditionally.

https://github.com/rosenpass/rosenpass/pull/521/commits/6b3bebb9b9b2326f8ce13a4a1b9e4be26b2576c2

Whereas previously, the futures crate was used in a block enabled on freebsd and linux exclusively

https://github.com/rosenpass/rosenpass/blob/c9da9b85913b8747e9965eaae30390f089cbec47/rp/src/exchange.rs#L35-L42

this PR introduced an unconditional import of the futures crate. The issues could be resolved by adding the futures crate to the list of dependencies, but since the futures crate was used only in the platform specific code block, I was able to resolve the issue without resorting to a non-platform specific code by adding appropriate platform feature flags to the pieces of code using the imports.

In other words, this was not a platform specific issue and merely a case of incorrectly configured platform specific code.

Can you please cherry-pick my fix into your branch (I can not push to yours) and rebase the branch?

@aparcar wanted to manually give this a whirl before merge, but other than that I think this PR is ready for merge.

Thank you for the contribution, this has become an excellent update to make `rp` more useful!

tfc:
@koraa i cherrypicked the commit. Please approve the PR workflows so that we can see how it works.

rp/src/cli.rs
```diff
@@ -144,6 +144,13 @@ impl ExchangeOptions {
                         return fatal("dev option requires parameter", Some(CommandType::Exchange));
                     }
                 }
+                "ip" => {
+                    if let Some(ip) = args.next() {
+                        options.ip = Some(ip);
+                    } else {
+                        return fatal("is option requires parameter", Some(CommandType::Exchange));
```
aparcar, line 156: This should be "ip option [...]" I reckon

tests/systemd/rp.nix
```diff
@@ -0,0 +1,139 @@
...
+      shared = peer: { config, modulesPath, pkgs, ... }: {
+        # Need to work around a problem in recent systemd changes.
+        # It won't be necessary in other distros (for which the systemd file was designed), this is NixOS specific
+        # https://github.com/NixOS/nixpkgs/issues/258371#issuecomment-1925672767
+        # This can potentially be removed in future nixpkgs updates
+        systemd.packages = [
+          (pkgs.runCommand "rp@.service" { } ''
+            mkdir -p $out/lib/systemd/system
+            < ${pkgs.rosenpass}/lib/systemd/system/rosenpass.target > $out/lib/systemd/system/rosenpass.target
+            < ${pkgs.rosenpass}/lib/systemd/system/rp@.service \
```
aparcar, line 64: Could you please explain these lines? Isn't it just a "fancy" `cp` and `rm`? If so I'd prefer to be more *verbose* and use those commands. Maybe a matter of taste. (same for `rosenpass.nix`)

tfc: Do you see the `sed` commands inbetween, that substitute all kinds of substrings? Can you please elaborate how you would you do that with `cp` and `rm`?

rp/src/cli.rs
tfc: fixed

tests/systemd/rp.nix
aparcar: Thanks, I misread that (still catching up on nix overall). Just for clarification, the first line only copies the contents right? 
```            
< ${pkgs.rosenpass}/lib/systemd/system/rosenpass.target > $out/lib/systemd/system/rosenpass.target
```

tfc: The first line just performs a plain copy, because in this file, no substitution happens.

The second file's content needs a lot of substitution.

Both lines simply happen in the "same style", which is of course a matter of taste.

However, this shell script snippet has nothing to do with nix inherently (apart from the ${...} string interpolation), it's pure bashism.

aparcar: Thanks!

tests/systemd/rosenpass.nix
```diff
@@ -0,0 +1,183 @@
...
+    secret_key = "/run/credentials/rosenpass@rp0.service/pqsk";
+    verbosity = "Verbose";
+    peers = [{
+      device = "rp0";
+      peer = client.wg.public;
+      public_key = "/etc/rosenpass/rp0/peers/client/pqpk";
+    }];
+  };
+  client_config = {
+    listen = [ "0.0.0.0:9999" ]; # TODO: Should not be necessary to set, but wouldn't parse.
```
aparcar, line 42:
```suggestion
    listen = [ ];
```

I just tested this and it works, too. Right now, `rosenpass` excepts it to be there, but if it's empty, [it performs some magic on it's own](https://github.com/rosenpass/rosenpass/issues/487).

aparcar:
I think this is really nice and especially the NixOS based configuration of two VMs running Rosenpass server/client quite impressive!

Since I barely touched NixOS prior to this, could you please give a minimal example on how this would look like in real life? Could you kindly give me some pointers? I got a VM running NixOS, how would I install Rosenpass with your changes (i.e. something in the lines of `nixos-install --flake https://github.com/rosenpass/rosenpass`?) and then configure it to connect to my hosts Rosenpass instance? I'd like to document this somehow since this seems super helpful for reproducible setups, however each NixOS guide I find suggests somewhat different steps (a fast moving target, I assume).

tests/systemd/rosenpass.nix
tfc: Thank you! Fixed.

tfc:
> I think this is really nice and especially the NixOS based configuration of two VMs running Rosenpass server/client quite impressive!
> 
> Since I barely touched NixOS prior to this, could you please give a minimal example on how this would look like in real life? Could you kindly give me some pointers? I got a VM running NixOS, how would I install Rosenpass with your changes (i.e. something in the lines of `nixos-install --flake https://github.com/rosenpass/rosenpass`?) and then configure it to connect to my hosts Rosenpass instance? I'd like to document this somehow since this seems super helpful for reproducible setups, however each NixOS guide I find suggests somewhat different steps (a fast moving target, I assume).

This workpackage has *nothing* to do with NixOS. The work order was to create systemd unit files that work on Ubuntu, Debian, Fedora, et. al. when they come packaged with the binary.

This is *not* the prefered way to set rosenpass or rp up on NixOS, as these already exist upstream in the nixpkgs repo.
The NixOS test solely proves that the new systemd unit files work well.
Tests that automate tests in legacy distros are planned but out of scope of this PR.

aparcar:
Thanks for this clarification, I'll merge this once all tests passed.

aparcar (approved):

koraa:
@aparcar Thank you! Good work!
@tfc You too of course.

@aparcar I am trying to start being consistent about how I merge PRs…I am now rebasing first and then I produce a merge commit that could be fast forward…just to document which commits belonged to which PR

PR #500, opened 2024-11-22
Add docstring examples and unit tests for the LengthPrefixEncoder module

phildremi:
~~Warning: Needs (a lot of) refinement. Placeholders abound.~~

---

Status: Should be ready for review/merge, pending further comments.

### DIFF
--- a/util/src/length_prefix_encoding/encoder.rs
+++ b/util/src/length_prefix_encoding/encoder.rs
@@ -121,6 +121,54 @@ pub struct WriteToIoReturn {
 
 #[derive(Clone, Copy, Debug)]
 /// An encoder for length-prefixed messages.
+///
+/// # Examples
+///
+///  ## Writing to output streams
+///
+/// Simplified usage example:
+///
+/// ```rust
+/// use rosenpass_util::length_prefix_encoding::encoder::LengthPrefixEncoder;
+/// use rosenpass_util::length_prefix_encoding::encoder::HEADER_SIZE;
+///
+/// let message = String::from("hello world");
+/// let mut encoder = LengthPrefixEncoder::from_message(message.as_bytes());
+///
+/// let mut output = Vec::new();
+/// encoder.write_all_to_stdio(&mut output).expect("failed to write_all");
+///
+/// assert_eq!(output.len(), message.len() + HEADER_SIZE);
+///
+/// let (header, body) = output.split_at(HEADER_SIZE);
+/// let length = u64::from_le_bytes(header.try_into().unwrap());
+///
+/// assert_eq!(length as usize, message.len());
+/// assert_eq!(body, message.as_bytes());
+/// ```
+///
+/// For more examples, see also:
+///
+/// * [Self::write_all_to_stdio]
+/// * [Self::write_to_stdio]
+///
+///
+/// ## Basic error handling
+///
+/// Creating an encoder with invalid parameters triggers one of the various sanity checks:
+///
+/// ```rust
+/// use rosenpass_util::length_prefix_encoding::encoder::{LengthPrefixEncoder, MessageLenSanityError};
+///
+/// let message_size = 32;
+/// let message = vec![0u8; message_size];
+///
+/// // The sanity check prevents an unsafe out-of-bounds access here
+/// let err = LengthPrefixEncoder::from_short_message(message, 2 * message_size)
+///	    .expect_err("OOB access should fail");
+/// assert!(matches!(err, MessageLenSanityError::MessageTooLarge(_)));
+/// ```
+
 pub struct LengthPrefixEncoder<Buf: Borrow<[u8]>> {
     buf: Buf,
     header: [u8; HEADER_SIZE],
@@ -145,6 +193,10 @@ impl<Buf: Borrow<[u8]>> LengthPrefixEncoder<Buf> {
     }
 
     /// Creates a new encoder using part of the buffer as a message
+    ///
+    /// # Example
+    ///
+    /// See [Basic error handling](#basic-error-handling)
     pub fn from_short_message(msg: Buf, len: usize) -> Result<Self, MessageLenSanityError> {
         let mut r = Self::from_message(msg);
         r.set_message_len(len)?;
@@ -159,12 +211,39 @@ impl<Buf: Borrow<[u8]>> LengthPrefixEncoder<Buf> {
     }
 
     /// Consumes the encoder and returns the underlying buffer
+    ///
+    ///	# Example
+    ///
+    /// ```rust
+    /// use rosenpass_util::length_prefix_encoding::encoder::LengthPrefixEncoder;
+    ///
+    /// let msg = String::from("hello world");
+    /// let encoder = LengthPrefixEncoder::from_message(msg.as_bytes());
+    /// let msg_buffer = encoder.into_buffer();
+    /// assert_eq!(msg_buffer, msg.as_bytes());
+    /// ```
     pub fn into_buffer(self) -> Buf {
         let Self { buf, .. } = self;
         buf
     }
 
     /// Consumes the encoder and returns buffer, message length and write position
+    ///
+    ///	# Example
+    ///
+    /// ```rust
+    /// use rosenpass_util::length_prefix_encoding::encoder::LengthPrefixEncoder;
+    ///
+    /// let msg = String::from("hello world");
+    /// let encoder = LengthPrefixEncoder::from_message(msg.as_bytes());
+    /// assert!(encoder.encoded_message_bytes() > msg.len());
+    /// assert!(!encoder.exhausted());
+    ///
+    /// let (msg_buffer, msg_length, write_offset) = encoder.into_parts();
+    /// assert_eq!(msg_buffer, msg.as_bytes());
+    /// assert_eq!(write_offset, 0);
+    /// assert_eq!(msg_length, msg.len());
+    /// ```
     pub fn into_parts(self) -> (Buf, usize, usize) {
         let len = self.message_len();
         let pos = self.writing_position();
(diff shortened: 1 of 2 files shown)


util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
 }
 
 #[derive(Clone, Copy, Debug)]
+/// Module description goes here (skipped since it already exists in #479)
```
koraa, line 95: Looks good for now! Lets use the coverage checker to make sure most of the functions from this file are covered.

Can you create a camera ready version and then write a list of possible additions in the comment in the PR?

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+///
+/// # Examples
+///
+///  Unfinished/proof-of-concept versions below - don't try this at home
+///
+/// **TODOs**:
+///
+/// * Actually write to a sink, if that's desirable/integration tests don't suffice
```
koraa, line 103: I think writing to a dummy sink (memory buffer, I think you can use std::io::Cursor along with static array) would be a great demo!

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// * Consider splitting into multiple examples in that case
```
koraa, line 104: You can, its not essential!

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// * Resolve all TBD notes, fixup/polish, add references/links, etc.
```
koraa, line 105: Not strictly required; don't polish too much. We want to cover most of the Repo!

koraa: Its more important to add some amount of documentation and – potentially – examples/tests to all objects in the repo rather than making sure that each individual test is 200%

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// * rustfmt ignores docstrings, check [unstable option](https://github.com/rust-lang/rustfmt/issues/3348) or fix manually?
```
koraa, line 106: We have a script for that  called `format_rust_code.sh`. If we could have rustfmt native support, that would be wonderful!

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// * Decide on the most important use cases vs. per-function examples (intended API usage?)
```
koraa, line 107: Lets settle on one use case, the integration test writing to a memory buffer above.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// * Unit test vs docstring test - unclear which approach is preferable here
```
koraa, line 108: Docstring test for the most part. Our goal is to have docstring tests that serve as good examples and cover most lines of code if possible.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+///
+/// ```
+/// use rosenpass_util::length_prefix_encoding::encoder::LengthPrefixEncoder;
+///
+/// let msg = String::from("hello world");
+/// let encoder = LengthPrefixEncoder::from_message(msg.as_bytes());
+/// assert!(encoder.encoded_message_bytes() > msg.len());
+/// assert!(!encoder.exhausted());
+///
+/// // TBD: should demonstrate several ways of creating new encoders?
```
koraa, line 122: Only insofar as you are making sure most functions in the file are being tested. No use getting too creative here.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// // TBD: relevant use cases to cover? need to mention trade-offs/perf/sec/other concerns?
```
koraa, line 123: Lets stick to the one discussed above.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+///
+/// // TBD: write_all_to_stdio -> assert that the entire message has been written (?)
```
koraa, line 125: How about just having a check in the integration test above ensuring that the right data was written.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// // TBD: write_to_stdio -> check chunk size/completion status after each step (?)
```
koraa, line 126: We can, but no need to get overly specific here. This is one we could skip and instead rely on the integration test above.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// ```
+/// use rosenpass_util::length_prefix_encoding::encoder::LengthPrefixEncoder;
+///
+/// let message = String::from("hello world");
+/// let mut encoder = LengthPrefixEncoder::from_message(message.as_bytes());
+///
+/// let mut output = Vec::new();
+/// encoder.write_all_to_stdio(&mut output).unwrap();
+///
+/// // TBD: Size of the header not part of the API, are exact checks needed?
```
koraa, line 147: Size of the header *is* part of the API.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+/// assert!(output.len() > message.len());
+///
+/// // TBD: usize impl detail (obv placeholder - highly questionable)
+/// let (header, body) = output.split_at(8);
+/// // TBD LE also internal -> decode_header ?
+/// let length = u64::from_le_bytes(header.try_into().unwrap());
```
koraa, line 153: I think relying on a reference output for a specific input would be fine.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+///
+/// ## Basic error handling
+///
+/// Creating a decoder with invalid parameters triggers one of the various sanity checks:
+///
+/// ```
+/// use rosenpass_util::length_prefix_encoding::encoder::{LengthPrefixEncoder};
+///
+/// let message = vec![0u8; 32];
+/// let result = LengthPrefixEncoder::from_short_message(message, 64); // TBD use vars for sizes(clarity)?
```
koraa, line 167: *shrug* I don't care. Lets work swiftly and not be too in-depth here.

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -92,6 +92,90 @@ pub struct WriteToIoReturn {
...
+///
+/// //TBD: not worth demonstrating since it's part of the signature? May still want to test regardless?)
+/// match result {
+/// // There are many Errors like it, but this one is mine :)
+///     Err(e) if matches!(e, MessageLenSanityError) => {
+///         println!("No bueno!"); // TBD: Probably irrelevant, delete it? Or test something else?
```
koraa, line 173: If it doesn't hurt anything it can stay :)

koraa (commented):
I think you did a really decent job here. Lets focus on increasing the coverage from here on. Nicely done!

koraa:
Relevant for #446

util/src/length_prefix_encoding/encoder.rs
phildremi: Coverage should be reasonably close to 100% now. The tool doesn't detect docstring tests and the final report doesn't show full coverage (78% for me), despite all functions being covered:

image

The coverage also improves when running only the docstring tests vs. running the entire test suite:

image

Since I've spent way too much time on trying to get more coverage already, I hope it will be enough.

util/src/length_prefix_encoding/encoder.rs
phildremi: It seems a bit out of scope, so maybe a separate issue would be warranted?

util/src/length_prefix_encoding/encoder.rs
phildremi: One of the examples/tests I added does something similar, inspired by the decoder tests.

util/src/length_prefix_encoding/encoder.rs
phildremi: I've used the exported constant in all examples/tests (unless I missed something).

util/src/length_prefix_encoding/encoder.rs
phildremi: It did hurt me, emotionally. That's why I've removed it in the final version :)

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -9,31 +18,31 @@ use zeroize::Zeroize;
 
 use crate::{io::IoResultKindHintExt, result::ensure_or};
 
-/// Size of the length prefix header in bytes - equal to the size of a u64
+/// Size in bytes of the message header carrying length information. Currently, HEADER_SIZE is always 8 bytes and encodes a 64-bit little-endian number.
```
aparcar, line 21: Please split this line

util/src/length_prefix_encoding/encoder.rs
```diff
@@ -185,7 +292,46 @@ impl<Buf: Borrow<[u8]>> LengthPrefixEncoder<Buf> {
         }
     }
 
-    /// Writes the next chunk of data to an IO writer and returns number of bytes written and completion status
+    /// Attempts to write the next chunk of data to an IO writer, returning the number of bytes written and completion flag
+    ///
+    /// # Example
+    ///
+    ///	```rust
+    ///
```
aparcar, line 300: Please drop this line.

util/src/length_prefix_encoding/mod.rs
```diff
@@ -1,4 +1,2 @@
-/// Module that handles decoding functionality
 pub mod decoder;
-/// Module that handles encoding functionality
 pub mod encoder;
```
aparcar, line 2: Why are those doc strings removed?

aparcar (changes requested):

util/src/length_prefix_encoding/encoder.rs
phildremi: Done!

aparcar (approved):

util/src/length_prefix_encoding/mod.rs
phildremi: @aparcar I'll consider this resolved for now :)

It would be possible to move the first line of the new summary into the `mod.rs` file, but it consists of two lines that you wouldn't look at when working on the docs, so that doesn't seem particularly useful here.

util/src/length_prefix_encoding/encoder.rs
phildremi: @koraa I've created a [separate tracking issue](https://github.com/rosenpass/rosenpass/issues/562). With that, I'll consider this point resolved :)

PR #501, opened 2024-11-22
Rename mio.connection.shoud_close (typo in function name)

phildremi:
Technically a breaking change... Hopefully that's not a problem here?


koraa:
Thank you!

PR #514, opened 2024-11-30
fix: allow multiple calls to `Oresp_hello` oracle

### LINKED ISSUE #358: ProVerif model should start Oresp_hello oracle with repetition
https://github.com/rosenpass/rosenpass/blob/19293471e85615a02c255afa910e9128b392b984/analysis/rosenpass/oracles.mpv#L189

The current model starts Oresp_hello just once as a continuation. Due to this, the model does not perfectly cover cases where the adversary calls the initiator with incorrect parameters. The initiator terminates its state after processing RespHello, but the oracle can be called multiple times until the first valid call.

surajk-m:
Closes #358

### DIFF
--- a/analysis/rosenpass/oracles.mpv
+++ b/analysis/rosenpass/oracles.mpv
@@ -186,7 +186,6 @@ let Oinitiator() =
   CK_EV( event OskOinitiator(ck, psk, sski, spkr, sptr); )
   MTX_EV( event IHSent(ih, psk, sski, spkr); )
   out(C, ih);
-  Oresp_hello(HS_PASS_ARGS).
 
 restriction sid:SessionId, ad1:Atom, ad2:Atom;
   event(ConsumeSidi(sid, ad1)) && event(ConsumeSidi(sid, ad2))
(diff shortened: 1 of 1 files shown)


analysis/rosenpass/oracles.mpv
```diff
@@ -205,7 +204,8 @@ let rosenpass_main() = 0
   | !Oreveal_kem_pk
   | REP(INITIATOR_BOUND, Oinitiator)
   | REP(RESPONDER_BOUND, Oinit_hello)
-  | REP(RESPONDER_BOUND, Oinit_conf).
+  | REP(RESPONDER_BOUND, Oinit_conf)
```
koraa, line 207: This gives me a syntax error when I try it locally.

koraa (changes requested):
Please make sure that the symbolic models actually compile and produce meaningful proofs of security.

koraa:
Looks like our CI currently doesn't properly run the proverif analysis which is why this error was hidden.

I just created an issue to fix that. #515

PR #528, opened 2024-12-11
Add documentation for the rp crate

DavidNiehues:
This PR adds documentation for the rp crate. It does not add tests since many unit tests already exist and in the areas where more tests are need, integration tests are the more suitable tool.


rp/src/exchange.rs
```diff
@@ -271,6 +309,7 @@ pub async fn exchange(options: ExchangeOptions) -> Result<()> {
         anyhow::Error::msg(format!("NativeUnixBrokerConfigBaseBuilderError: {:?}", e))
     }
 
+    // configure everything per peer
```
aparcar, line 312: A style thing, should we always start comments uppercase or keep it random? CC @koraa

DavidNiehues: Is there a consistent policy in the other crates? If so, I'm happy to follow that policy as well.

Either way, it shouldn't be too much of an issue for me to change the hand full of comments to match what is done in the other crates.

aparcar (approved):

PR #535, opened 2024-12-14
WP2: Create DEB&RPM packages and test in debian/ubuntu/fedora

tfc:
This PR does:

- generate `.rpm` and `.deb` packages which contain:
  - the statically linked binaries
  - systemd unit files
  - example configuration file
- test the .rpm and .deb packages on Debian, Ubuntu, fedora:
  - by copying and editing the `example.toml` config file that came with the package
  - launches two `rp` service instances on the same host and checks if their PSK exchange works properly

[Numtide's `nix-vm-test` library](https://github.com/numtide/nix-vm-test) that generates Debian/Ubuntu/Fedora VMs is used to to the legacy distro testing.
It comes with the current limitation that it only supports single-host tests.
This is the reason why the tests don't spin up a VPN connection over network between two hosts.

Personally,  i consider this fine from a testing perspective because we only need to test if the package installation&configuration work, which it currently does very well.
Other tests check the general functionality of rosenpass and rp.

cc @koraa


pkgs/package-deb.nix
```diff
@@ -0,0 +1,30 @@
...
+in
+
+runCommand "rosenpass-${version}.deb" { } ''
+  mkdir -p packageroot/DEBIAN
+
+  cat << EOF > packageroot/DEBIAN/control
+  Package: rosenpass
+  Version: ${version}
+  Architecture: all
+  Maintainer: Jacek Galowicz <jacek@galowicz.de>
```
tfc, line 14: Not sure who would be the appropriate maintainer in this field. Please advise.

flake.nix
```diff
@@ -150,7 +165,11 @@
               { nativeBuildInputs = [ pkgs.nodePackages.prettier ]; } ''
               cd ${./.} && prettier --check . && touch $out
             '';
-          };
+          } // pkgs.lib.optionalAttrs (system == "x86_64-linux") (import ./tests/legacy-distro-packaging.nix {
```
tfc, line 168: I'm limiting the testing to x86_64 linux at this point because i don't have an ARM machine with KVM available right now.

The tests should be simple to adapt to aarch64 (they would have to obtain arm wireguard packages to support that).

tests/legacy-distro-packaging.nix
```diff
@@ -0,0 +1,71 @@
+{ pkgs, rosenpass-deb, rosenpass-rpm }:
+
+let
+  wg-deb = pkgs.fetchurl {
+    url = "https://ftp.de.debian.org/debian/pool/main/w/wireguard/wireguard-tools_1.0.20210914-1.1_amd64.deb";
+    hash = "sha256-s/hCUisQLR19kEbV6d8JXzzTAWUPM+NV0APgHizRGA4=";
+  };
+  wg-rpm = pkgs.fetchurl {
+    url = "https://mirrors.n-ix.net/fedora/linux/releases/40/Everything/x86_64/os/Packages/w/wireguard-tools-1.0.20210914-6.fc40.x86_64.rpm";
```
aparcar, line 9: Does this need to be tracked and updated from time to time?

tfc: I think the link will work for a very long time.

The necessity to upgrade the rpm and deb packages might possibly come up whenever the distros are updated.

Upgrading would look like this:

1.) One would update the `nix-vm-test` input and re-run the tests.
2.) In the worst case, the packages wouldn't work in the tests any longer for whatever reason
3.) The updater would put updated URLs and hashes in the fetcher calls.

The whole process would take a few minutes and be performed max once a year if not less often - i wouldn't advise to invest the time for any automation here.

aparcar:
This fails on a aarch64-linux machine:

image

tfc:
> This fails on a aarch64-linux machine:
> 
> image

That's interesting because this job has nothing to do with the test, and it also went through green in the CI. Are you sure that your git index was unchanged relative to the commits?

aparcar:
I'm assuming an OOM in my build environment, will let you know in a bit.

pkgs/package-rpm.nix
```diff
@@ -0,0 +1,57 @@
...
+  splitVersion = lib.strings.splitString "-" rosenpass.version;
+  version = builtins.head splitVersion;
+  release =
+    if builtins.length splitVersion != 2
+    then "release"
+    else builtins.elemAt splitVersion 1;
+  arch = builtins.head (builtins.split "-" system);
+in
+
+runCommand "rosenpass-${version}.deb" { } ''
```
aparcar, line 13:
```suggestion
runCommand "rosenpass-${version}.rpm" { } ''
```

aparcar:
Turns out to be an OOM issue, sorry for the noise.

It worked and I got a result symlink which would link in for both DEB and RPM to a file named <>.deb, I commented the typo above.

Would it be possible to use Nix to create a more meaningful name than `result`?

tfc:
> Would it be possible to use Nix to create a more meaningful name than `result`?

It's possible, but you have to do that yourself by running `nix build .#bla -o my-out-link`. The Nix expressions cannot be modified in a way that they would create something else automatically. But this is also not how you would consume the output normally.

aparcar:
May I asked how this is done normally?

tfc:
> I may mistake the idea of the WP but from what I understand it's about offering DEB and RPM packages, which could be installed downstream. Shouldn't we offer DEB/RPM over a signed repository so downstream users can make real use of it?

Absolutely. It should be served from a deb/rpm package mirror that can be added to the APT/DNF/YUM package manager config of a user, along with their keys, so that users can simply `apt/dnf install rosenpass`.

But this PR has nothing to do with that. The work package is solely about generating the packages and testing them.

aparcar:
Awesome, I think this PR is mostly ready! Thanks for your work on this, really good progress forward to make this usable for more people!

Please add as final step an [upload-artifact](https://github.com/actions/upload-artifact) job so it's possible to test the resulting artifacts outside of the Nix testing environment.

Generally we try to encourage commits to follow the [semantic commit message schema](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716), however if you're unfamiliar with `git rebase -i` or low on time, please leave the commits as they are.

tfc:
@aparcar the latest commit should create the upload artefact step you asked for, can you have a look?

koraa:
@tfc Thank you for the contribution!

PR #538, opened 2024-12-15
docs(util): fix doc reference in decoder.rs

aminfa:
docs(util): add more tests and example to complete coverage


aparcar:
Thanks for your work! Commit wise you'd ideally add two commits instead of squashing them, due to this size of the commit I think it doesn't matter this time.

aparcar (approved):

PR #542, opened 2024-12-15
Fix doctests where Function is wrapped around the actual test but is never called in cipher traits

DavidNiehues:
In the doctests in `kem.rs`, the actual tests that are run to verify that the KyberKem and the DummyKem actually work
are wrapped inside a function to make use of the `?`-operator. However, these functions were never called and thus the tests weren't really helpful and didn't provide proper coverage. This PR fixes this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1866,6 +1866,7 @@ dependencies = [
 name = "rosenpass-cipher-traits"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "rosenpass-oqs",
  "rosenpass-secret-memory",
 ]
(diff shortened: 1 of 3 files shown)


aparcar:
Please add the PR message to the git commit message.
Do you know why the wrapper is used after all? CC @koraa

DavidNiehues:
> Please add the PR message to the git commit message. Do you know why the wrapper is used after all? CC @koraa
Yes, the example uses the `?`-operator. That requires that the code is part of a function that returns as `Result`. That is what the function wrapping the test achieves.

koraa:
@DavidNiehues Can you adjust the PR to the above pattern, then rebase it?

@aparcar Do  you approve this being merged?

koraa:
@DavidNiehues Please run "cargo fmt" again

DavidNiehues:
@koraa Just now saw your last comment. Should already be fixed.

PR #544, opened 2024-12-16
docs(constant-time): add docs, examples and safety notices

aminfa:


aparcar:
Looks good to me, I'll see why the QC CI fails.

@koraa are you aware of constant-time implementations which don't have the stated drawbacks?

aparcar (approved):

constant-time/src/compare.rs
```diff
@@ -2,14 +2,28 @@
...
+/// Little endian memcmp version of [quinier/memsec](https://github.com/quininer/memsec/blob/bbc647967ff6d20d6dccf1c85f5d9037fcadd3b0/src/lib.rs#L30)
 ///
 /// # Panic & Safety
 ///
 /// Both input arrays must be at least of the indicated length.
 ///
 /// See [std::ptr::read_volatile] on safety.
+///
+/// # Examples
+/// ```no_run
```
aparcar, line 14: I think this fails to compile since it does not find `memcmp_le`. I think you should either import it via something like `use rosenpass_constant_time::memcmp_le` or replace `no_run` with `ignore`.

koraa:
> Looks good to me, I'll see why the QC CI fails.
> 
> @koraa are you aware of constant-time implementations which don't have the stated drawbacks?

Do you mean these?

> //! - Length mismatches between inputs are immediately detectable
> //! - Execution time scales linearly with input size

No I am not aware of alternatives

constant-time/src/compare.rs
aminfa: I am not able to import `memcmp_le` because the module `compare` is private.

I can move the test into a unit-test instread and add the `ignore` attribute instread

koraa: @aminfa You can also just make the compare module public if that is appropriate. We generally have an "everything is public but no API is stable" policy at the moment.

aminfa: Ok, I exported memcmp_le as a public item and removed the `ignore` doc-test attribute

aminfa:
The QC fails seem to stem from the statistical tests in `memcmp_runs_in_constant_time`. As I haven't touched any code, this seems to be a false positive

PR #545, opened 2024-12-16
docs(util): add docs and examples for the remaining util crate

aminfa:


util/src/build.rs
```diff
@@ -508,9 +508,9 @@ where
         matches!(self, Self::Void)
     }
 
-    /// Returns `true` if the construction site is [`InProgress`].
+    /// Returns `true` if the construction site is in the [`Builder`] phase.
     ///
-    /// [`InProgress`]: ConstructionSite::InProgress
+    /// [`Builder`]: ConstructionSite::Builder
```
aparcar, line 513: Nice Catch

util/src/mem.rs
```diff
@@ -1,10 +1,32 @@
...
+//! and various traits and types that help manage values, including preventing
+//! drops, discarding results, and swapping values.
+
 use std::borrow::{Borrow, BorrowMut};
 use std::cmp::min;
 use std::mem::{forget, swap};
 use std::ops::{Deref, DerefMut};
 
-/// Concatenate two byte arrays
-// TODO: Zeroize result?
```
aparcar, line 7: Does this happen now somewhere?

util/src/mem.rs
```diff
@@ -150,11 +245,16 @@ impl<T: Default> SwapWithDefaultExt for T {
 }
 
 /// Extension trait that provides a method to explicitly move values.
+///
+/// # Examples
+///
+/// ```rust
+/// use rosenpass_util::mem::MoveExt;
+/// let val = 42.move_here();
```
aparcar, line 253: Is there a better example for this? To me it looks like `let val = 42; assert_eq!(val, 42);` return fine, too.

aparcar:
Thanks for the PR, overall I think it's all very fine, can't wrap my head around the typenum stuff but should be good, too.

util/src/mem.rs
aminfa: Sorry, I accidently removed the TODO. I will re-insert it

util/src/mem.rs
aminfa: I think in Rust you can't directly "test" if a value has been moved because once a value is moved, it becomes inaccessible. 

I can add an additional comment that explains that the initial variable will be inaccessible. 

Alternatively, I can test if the value is dropped by using a custom type that signals as soon as it is dropped (setting an AtomicBool for example). But I think that that would be an "overkill" example and I can instead add it as a unittest.

aparcar: Okay if you think this is common knowledge I'm fine with it.

util/src/mem.rs
```diff
@@ -150,11 +246,28 @@ impl<T: Default> SwapWithDefaultExt for T {
...
+/// assert_eq!(Rc::strong_count(&value), 1);
+/// ```
 pub trait MoveExt {
-    /// Deliberately move the value
-    ///
-    /// Usually employed to enforce an object being
-    /// dropped after use.
+    /// Move `self` here.
     fn move_here(self) -> Self;
 }
```
aminfa, line 272: I hope this is a better Example

koraa:
@aparcar Can you merge this when ready?

aparcar (approved):

PR #557, opened 2024-12-18
Switch from Kyber to mlkem.

loganaden:
Joint work with Pratish Sungum.

Sign-off-by: Loganaden Velvindron <logan@cyberstorm.mu>


aparcar:
Thanks for your contribution, please add some context to this and write it into the commit message.

koraa:
Thanks for updating the liboqs version here.

This is a breaking change; the main requirement for updating to liboqs is that we introduce a configuration flags to fall back to the old algorithms, allowing established users of rosenpass to maintain compatibility with existing clients.

Would you be up for giving it a try? If yes, then we should talk about the design of the configuration flag.

loganaden:
@koraa sure. How do you propose doing that ?

ilka-schulz:
This would be a breaking change and requires re-planning. Sorry for the long delay. I will put this topic on my todo list and come up with some planning.

PR #577, opened 2025-01-14
Add providers and more crypto traits

keks:
# Rosenpass PR Text Draft

This PR adds more traits for primitives, as well as a provider abstraction for swapping out the crypto implementation that is being used.
Please see it as a conversation starter, not a finished product. There are some things that I think could be better about this, but it’s not my code base so I’d like to get feedback early.
First, a brief rundown of the changes:

* Added traits for remaining primitives
  * Parameters are const generics, not associated constants
    * This is more verbose, but given the limitations around associated constants and const expressions, it seems to me that this way we can use the constants a lot better
* Add marker traits for algorithms
  * E.g. `ChaCha20Poly1305`, which is a subtrait of `Aead<32, 32, 12, 16>` and implemented in both providers
* Provider trait with associated types for all algorithms
  * Implemented both by rust crypto and libcrux
    * Libcrux falls back to rustcrypto in some cases
  * Trait is stateless, all functionality lives in associated functions (without self arguments)
* Depending on set feature flags, `rosenpass_ciphers::Provider` is either `libcrux::Provider` or `basic::Provider` (via a conditional use)
* Code in actual rosenpass uses `rosenpass_ciphers::Provider` (via indirections)

Two things are not quite finished yet:

- The libcrux provider does not yet use libcrux for blake2, but falls back on the other provider. I am currently adding pure-rust blake2 to libcrux. Once that is done, I will port it to that.
- Currently, this uses libcrux from a git commit instead a released version. This is because some of the APIs this uses are not part of a release yet. We are planning on doing another pre-release to crates.io soon, likely after blake is in.

These clearly need to get done before this gets merged, but I thought it would beer good to get feedback on the general approach sooner rather than later.

Next, let’s go over the things that I think need particular discussion.

# Just one Provider, determined at compile time

The user code just uses the implementation of the provider trait that has been configured at compile time, instead of having the protocol code be generic over the provider. From a “looking at the requirements” perspective I think this is fine, and from a code readability standpoint it’s also not bad because the user code is not cluttered with `<P: Provider>` noise. But it does feel like we are not using traits properly.

Note that the provider is fully stateless and we only use it as a type, not as a value. So as it stands, we need to explicitly spell it out as a generic argument. We could change and pass an empty struct around, just so the type inference works. That would also allow us to introduce state later (e.g. for randomness).

Questions
- Should we make the protocol implementation generic over the provider or keep using a provider struct at a well-known modpath?
- Should we pass around provider values or keep using just the type?

# Indirections
The `rosenpass-ciphers` crate has a module that contains functions that just call the provider, and the user code uses these functions. This means that there is no mention of providers at all in the user code, but it does feel like we didn’t finish refactoring.

It seems like all the things in `subtle` should now live inside the providers, so in theory we should be able to drop this directory. In fact, all these functions do is call into the provider, so maybe that is a layer of indirection that should be removed.

At the root of the `ciphers` crate there are submodules `aead`, `xaead`, `keyed_hash` and `kem`, which closely matches the “building blocks” from the whitepaper. This is good, the only comment I have is that they are not structured uniformly; the `kem` module has two structs with associated functions, whereas the other modules just contain functions. Maybe instead of `kem` we should have `static_kem` and `ephemeral_kem` with just a few functions inside.

Questions:
- Should we get rid of the subtle directory?
- Should we make the modules structured more uniformly? How? By making everything more “traity” (like kem is now) or more “functions-in-modules”-y (which aead and xaead are now)?
- My opinion: It seems to me that this layer is for removing all generics and just calling the functions of the chosen provider. That’s why I’d make this as un-traity as possible and just go with functions in modules.

## Bonus Questions

These are not specific to the changes here but are things that came up while working on the code and I just wondered about:

- Is (the somewhat weird) HMAC-Blake2 going to stick around? Was that ever a good idea? Blake2 should be a PRF already (and avoids the problem that resulted in the need for HMAC to begin with).
- Should Kyber be replaced by ML-KEM? Longer keys?


rosenpass/src/protocol/protocol.rs
```diff
@@ -3753,7 +3785,10 @@ impl CryptoServer {
                 let cookie_key = hash_domains::cookie_key()?.mix(spkt)?.into_value();
                 let cookie_value = peer.cv().update_mut(self).unwrap();
 
-                xaead::decrypt(cookie_value, &cookie_key, &mac, &cr.inner.cookie_encrypted)?;
+                let (nonce, ciphertext) = cr.inner.cookie_encrypted.split_at(24);
```
koraa, line 3788: Is having this here correct?

koraa (commented):
Notes from convo:

ciphers::subtle::oqs::...
ciphers::subtle::rust_crypto::...
ciphers::subtle::libcrux::...

ciphers::backend::default
ciphers::backend::libcrux

backend
------
subtle
    

- How to actually provide specific providers at runtime (towards)
- Adding more to the libcrux privider
- Making things easier with respect

keks:
This has been the bases of later work in #617, which then went into #625.

PR #581, opened 2025-01-28
feat(ciphers+fuzz): Add experimental feature to use a keyed version of SHAKE256 instead of Blake2b.

DavidNiehues:
This PR adds support for using a keyed version of SHAKE256, truncated to 32 bytes, instead of the incorrect HMAC with Blake2b as the PRF for key derivation for rosenpass.

This feature is a compile time flag as of now, namely the flag "experiment_sha3".

Since there is also fuzzing for Blake2b, this PR also disables this fuzzing if the feature "experiment_sha3" is disabled.

### DIFF
--- a/ciphers/src/subtle/keyed_shake256.rs
+++ b/ciphers/src/subtle/keyed_shake256.rs
@@ -1,0 +1,65 @@
+use anyhow::ensure;
+use rosenpass_to::ops::copy_slice;
+use rosenpass_to::{with_destination, To};
+use sha3::digest::{ExtendableOutput, Update, XofReader};
+use sha3::Shake256;
+use zeroize::Zeroizing;
+
+/// The key length, 32 bytes or 256 bits.
+pub const KEY_LEN: usize = 32;
+/// The output length, 32 bytes or 256 bits.
+pub const OUT_LEN: usize = 32;
+/// The minimal key length, identical to [KEY_LEN].
+pub const KEY_MIN: usize = KEY_LEN;
+/// The maximal key length, identical to [KEY_LEN].
+pub const KEY_MAX: usize = KEY_LEN;
+/// The minimal output length.
+pub const OUT_MIN: usize = OUT_LEN;
+/// The maximal output length.
+pub const OUT_MAX: usize = OUT_LEN;
+
+/// Provides a keyed hash function based on SHAKE256. To work for the protocol, the output length
+/// and key length are fixed to 32 bytes (also see [KEY_LEN] and [OUT_LEN]).
+///
+/// Note that the SHAKE256 is designed for 64 bytes output length, which we truncate to 32 bytes
+/// to work well with the overall protocol. Referring to Table 4 of FIPS 202, this offers the
+/// same collision resistance as SHAKE128, but 256 bits of preimage resistance. We therefore
+/// prefer a truncated SHAKE256 over SHAKE128.
+///
+/// #Examples
+/// ```rust
+/// # use rosenpass_ciphers::subtle::keyed_shake256::hash;
+/// use rosenpass_to::To;
+/// let key: [u8; 32] = [0; 32];
+/// let data: [u8; 32] = [255; 32];
+/// // buffer for the hash output
+/// let mut hash_data: [u8; 32] = [0u8; 32];
+///
+/// assert!(hash(&key, &data).to(&mut hash_data).is_ok(), "Hashing has to return OK result");
+/// # let expected_hash: &[u8] = &[166, 35, 212, 217, 225, 226, 193, 131, 163, 196, 223, 79, 56,
+/// 193, 107, 23, 45, 213, 14, 86, 198, 177, 49, 182, 233, 217, 157, 39, 188, 240, 140, 163];
+/// # assert_eq!(hash_data, expected_hash);
+/// ```
+pub fn hash<'a>(key: &'a [u8], data: &'a [u8]) -> impl To<[u8], anyhow::Result<()>> + 'a {
+    with_destination(|out: &mut [u8]| {
+        // Since SHAKE256 is a XOF, we fix the output length manually to what is required for the
+        // protocol.
+        ensure!(out.len() == OUT_LEN);
+        let mut out: [u8; OUT_LEN] = out.try_into().unwrap();
+        // Not bothering with padding; the implementation
+        // uses appropriately sized keys.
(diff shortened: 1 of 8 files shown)


koraa:
@DavidNiehues Can you rebase the pull request? Its set to prevent me from doing so.

fuzz/fuzz_targets/blake2b.rs
```diff
@@ -1,18 +1,18 @@
-#![no_main]
+# ![no_main]
```
koraa, line 1: I think this file needs to be renamed then. No need to do this right now, but I think it makes sense to allow both the sha3 and broken blake2b hash functions to coexist (feature flag "hash-blake2b", "hash-sha3") to allow things like fuzzing and tests to use them directly

ciphers/src/subtle/keyed_shake256.rs
```diff
@@ -0,0 +1,63 @@
...
+        let mut shake256 = Shake256::default();
+        shake256.update(key);
+        shake256.update(data);
+
+        let mut tmp = Zeroizing::new([0u8; OUT_LEN]);
+        // Following the NIST recommendations in Appendix A.2 of the FIPS 202 standard,
+        // we append the length of the input to the end of the input.
+        shake256.update(&(32u32.to_be_bytes()));
+        shake256.finalize_xof().read(tmp.as_mut());
+        copy_slice(tmp.as_ref()).to(out);
```
koraa, line 60: Can we get rid of the temp variable? Copy_slice needs to assert anyway so we might also use try_into to just cast the input to a `&mut [u8; 32]` to feed it into finalize directly.

ciphers/src/subtle/keyed_shake256.rs
```diff
@@ -0,0 +1,63 @@
...
+        // uses appropriately sized keys.
+        ensure!(key.len() == KEY_LEN);
+        let mut shake256 = Shake256::default();
+        shake256.update(key);
+        shake256.update(data);
+
+        let mut tmp = Zeroizing::new([0u8; OUT_LEN]);
+        // Following the NIST recommendations in Appendix A.2 of the FIPS 202 standard,
+        // we append the length of the input to the end of the input.
+        shake256.update(&(32u32.to_be_bytes()));
```
koraa, line 58: Why big endian? Is this the length of the key?

ciphers/src/subtle/keyed_shake256.rs
```diff
@@ -0,0 +1,63 @@
...
+        ensure!(out.len() == OUT_LEN);
+        // Not bothering with padding; the implementation
+        // uses appropriately sized keys.
+        ensure!(key.len() == KEY_LEN);
+        let mut shake256 = Shake256::default();
+        shake256.update(key);
+        shake256.update(data);
+
+        let mut tmp = Zeroizing::new([0u8; OUT_LEN]);
+        // Following the NIST recommendations in Appendix A.2 of the FIPS 202 standard,
```
koraa, line 56: Good, but then I think we need to make this part of the name of the module.

Is there any particular reason for this recommendation by the way? Sounds somewhat like cruft but I do not know.

koraa (changes requested):
Looks good for the most part, minor comments then LGTM

ciphers/src/subtle/keyed_shake256.rs
```diff
@@ -0,0 +1,63 @@
...
+pub const KEY_MIN: usize = KEY_LEN;
+/// The maximal key length, identical to [KEY_LEN].
+pub const KEY_MAX: usize = KEY_LEN;
+/// The minimal output length.
+pub const OUT_MIN: usize = OUT_LEN;
+/// The maximal output length.
+pub const OUT_MAX: usize = OUT_LEN;
+
+/// Provides a keyed hash function based on SHAKE256. To work for the protocol, the output length
+/// and key length are fixed to 32 bytes (also see [KEN_LEN] and [OUT_LEN]).
```
phildremi, line 22:
```suggestion
/// and key length are fixed to 32 bytes (also see [KEY_LEN] and [OUT_LEN]).
```

ciphers/src/subtle/keyed_shake256.rs
DavidNiehues: Sure thing. Not sure why I didn't do it that way directly.

ciphers/src/subtle/keyed_shake256.rs
DavidNiehues: It's the output length and thus should also use the `OUT_LEN` constant. The encoding in big endian is arbitrary, it could be any other reasonable encoding of the output length just as well.

ciphers/src/subtle/keyed_shake256.rs
DavidNiehues: Good find!

ciphers/src/subtle/keyed_shake256.rs
DavidNiehues: **Why:** The reason to include this is what @phildremi quoted. Since SHAKE256 is a XOF, two outputs with different lengths but the same input will be identcal up to the length of the shorter output, which can lead to related key attacks. I am not aware of any vulnerability that would be introduced by not dong this given that rosenpass does use domain separation and only uses a fixed key and output length of 32 bytes. I added it as a safeguard to make the implementation more robust with regards to future changes. If you prefer it that way, we can of course do without incorporating the output length.

*Appendix  A.2** Indeed, while A.x, B.x, ... usually refers to an appendix, everywhere in the standard, it is referred to as a section. I'll thus update the reference.

fuzz/fuzz_targets/blake2b.rs
DavidNiehues: Good point, at least the file name is misleading as of now.

ciphers/src/subtle/keyed_shake256.rs
koraa: Lets use little endian then; we generally use little endian in most places in rosenpass.

ciphers/src/subtle/keyed_shake256.rs
koraa: @phildremi Thank you for the help!

ciphers/src/subtle/keyed_shake256.rs
DavidNiehues: Done

koraa:
Already merged through another PR #631

PR #582, opened 2025-01-28
[WIP] feat(docker): add Dockerfile, docker-build script and docker README

aminfa:
I have added a Dockerfile with two final stages: one image for rosenpass and one for rp.
I have added  a corresponding build script that passes the args for the OCI labels.  
I have also added a README for how to build and use the docker images. 

I think this should cover both stand-alone rosenpass setups and also all-in-one (wg+rosenpass) setups, nicely. 

I have added annotations to the Docker image: [OpenContainers Image Spec - Annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md).

The purpose of these metadata entries is:
- Documentation: Labels standardize the description of the image's purpose, author, and version.
-  Automation: They facilitate identification and management in CI/CD pipelines and Kubernetes.
- Compliance: Labels store license and audit-relevant information for users who require this.

**Could you briefly review this and check if my entries are correct?**

As for the license, I specified “MIT OR Apache-2.0,” but if the project uses a multi-license model, it should be “MIT AND Apache-2.0.”. **I couldn’t determine which model applies to Rosenpass. **

Currently, the build script allows to create release tags, and it derives the version from the version specified in the corresponding `cargo.toml` file. (e.g. `rp/cargo.toml`). **Is this the correct way to label the images for releases? **

Let me know if I can improve anything.


koraa:
@aminfa Does it make sense to move most of these files into a subdirectory?

You also need to rebase this branch (its set to prevent me from triggering the rebase myself) and there are true-positive CI failures which need fixing.

Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+WORKDIR /app
+COPY . .
+RUN cargo build --release
+
+
+# Stage 5: Download tini and check it with a static sha256sum (to prevent supply-chain attacks)
+FROM debian AS with_tini
+RUN apt-get update && apt-get install -y wget && rm -rf /var/lib/apt/lists/*
+ENV TINI_VERSION=v0.19.0
+RUN wget --no-check-certificate --no-cookies --quiet https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-amd64 \
```
aparcar, line 34: This limits the containers to amd64 while Rosenpass containers would run perfectly fine on i.e. arm64.

docker-buildscript.sh
```diff
@@ -0,0 +1,55 @@
...
+    local VCS_REF=$(git rev-parse HEAD)
+    local VSC_SHORT_REF=$(git rev-parse --short HEAD)
+
+    local IMAGE_TAG="ghcr.io/utilacy/${TARGET}:commit-${VSC_SHORT_REF}"
+
+    docker build \
+        --target "$TARGET" \
+        --build-arg BASE_IMAGE=${BASE_IMAGE} \
+        --build-arg BASE_DIGEST=${BASE_DIGEST} \
+        --build-arg BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
```
aparcar, line 24: Should we use [SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/) here and set it in the CI to the date of the last commit, to allow reproducibility?

docker-buildscript.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Set the default base image if it is not set
+: "${BASE_IMAGE:="debian:bookworm-slim"}"
+
+TARGET=rosenpass
+DESCRIPTION="The reference implementation of the protocol - the rosenpass tool"
+build_docker_image
+
+TARGET=rp
+DESCRIPTION="A frontend integrating Rosenpass and WireGuard to create a vpn - the rp frontend"
```
aparcar, line 54:
```suggestion
DESCRIPTION="A frontend integrating Rosenpass and WireGuard to create a VPN - the rp frontend"
```

docker-buildscript.sh
```diff
@@ -0,0 +1,55 @@
...
+    local BASE_DIGEST=$(docker inspect --format='{{index .RepoDigests 0}}' ${BASE_IMAGE})
+    # Extract the digest part (after the @ symbol)
+    local BASE_DIGEST=${BASE_DIGEST#*@}
+
+    
+
+    local VCS_REF=$(git rev-parse HEAD)
+    local VSC_SHORT_REF=$(git rev-parse --short HEAD)
+
+    local IMAGE_TAG="ghcr.io/utilacy/${TARGET}:commit-${VSC_SHORT_REF}"
```
aparcar, line 18:
```suggestion
    local IMAGE_TAG="ghcr.io/rosenpass/${TARGET}:commit-${VSC_SHORT_REF}"
```

docker-buildscript.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
aparcar, line 1: Generally, is there a requirement to have this script? I'd run this from a CI anyway and therefore use the GitHub actions like [metadata-action](https://github.com/docker/metadata-action) and [build-push-action](https://github.com/docker/build-push-action).

readme.docker.md
```diff
@@ -0,0 +1,241 @@
...
+mkdir ./workdir
+docker run -it --rm -v ./workdir:/workdir ghcr.io/rosenpass/rosenpass \
+    gen-keys --public-key=workdir/server-public --secret-key=workdir/server-secret
+docker run -it --rm -v ./workdir:/workdir ghcr.io/rosenpass/rosenpass \
+    gen-keys --public-key=workdir/client-public --secret-key=workdir/client-secret
+```
+
+> Note: in this example the keys are accessible by both containers as they lie on the same host.
+> In practice, it is neccessary to make the public keys of the client and the server available to each other.
+> **DO NOT** publish or share the secret keys and keep them safe.
```
aparcar, line 81:
```suggestion
```bash
mkdir ./workdir-client ./workdir-server
docker run -it --rm -v ./workdir-server:/workdir ghcr.io/rosenpass/rosenpass \
    gen-keys --public-key=workdir/server-public --secret-key=workdir/server-secret
docker run -it --rm -v ./workdir-client:/workdir ghcr.io/rosenpass/rosenpass \
    gen-keys --public-key=workdir/client-public --secret-key=workdir/client-secret
 cp workdir-client/client-public workdir-server/client-public
 cp workdir-server/server-public workdir-client/server-public
```

> **DO NOT** publish or share the secret keys and keep them safe.
```

I'd walk the extra mile and not suggest a bad setup

**Review by aparcar — CHANGES_REQUESTED** [review]
Hi, nice work, thanks for your contribution. I added some comments inline, my biggest questions is how this related to our CI building it's own OCI containers via NixOS (`nix build .#rosenpass-oci-image`). Those are not published but we should come up with one solution allowing users to just pull ready containers.

If we want to use the script instead of GitHub actions, I'd move it into ./docker/ and rename the readme to ./docker/README.md. Either way, I'd like to see CI creating usable containers.

File: docker-buildscript.sh (hunk 3 above)
Comment (phildremi) [reply docker-buildscript.sh:54]: Alternative suggestions/phrasing:

```suggestion
DESCRIPTION="The rp frontend integrates Rosenpass and WireGuard to create a VPN"
```

```suggestion
DESCRIPTION="The rp command-line integrates Rosenpass and WireGuard to help you create a VPN"
```

```suggestion
DESCRIPTION="Command-line application integrating Rosenpass and WireGuard to create a VPN - the rp frontend"
```

```suggestion
DESCRIPTION="rp is a frontend integrating Rosenpass and WireGuard to help you create a VPN"
```

(Or some variation thereof. You could also consider replacing `create` with `set up` or even `manage`?)

.dockerignore
```diff
@@ -0,0 +1,29 @@
+# Rust
```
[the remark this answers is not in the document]
aminfa: We can do this. Generally it could be potentially dangerous as .dockerignore patterns are not exactly the same as .gitignore patterns: https://zzz.buzz/2018/05/23/differences-of-rules-between-gitignore-and-dockerignore/

Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+RUN cargo chef cook --release --recipe-path recipe.json
+
+# Stage 4: Build the application
+FROM cacher AS builder
+WORKDIR /app
+COPY . .
+RUN cargo build --release
+
+
+# Stage 5: Download tini and check it with a static sha256sum (to prevent supply-chain attacks)
```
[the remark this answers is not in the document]
aminfa: In my testing, SIGKILL did not properly kill all processes started by rp. Or at least that's what I think happened. That is why I added tini to handle signal forwarding and process reaping, which can help with proper shutdown and avoiding zombie processes. However, if rosenpass and rp runs as a single process and already handles SIGTERM correctly, we might not need it.

aminfa:
> Does it make sense to move most of these files into a subdirectory?

If we don’t need multiple Dockerfiles, keeping everything in the root directory is simpler and more conventional, as it aligns with common practices (tools and CI/CD systems typically expect the Dockerfile in the root). It also means we can run docker build . without needing to specify a path. 

I guess it is a matter of taste, and you may decide if you’d rather like a more clean root directory or if you’d prefer to stick to conventions.

docker-buildscript.sh
aminfa: I wanted to  conform to the [OCI image spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md#:~:text=org.opencontainers.image.created%20date%20and%20time%20on%20which%20the%20image%20was%20built%2C%20conforming%20to%20RFC%203339.).

But if you prefer SOURCE_DATE_EPOCH we can use that

docker-buildscript.sh
aminfa: I think you are right: if the main the images are built are with github actions then we should use the official github actions from docker. I will remove this build script.

readme.docker.md
aminfa: Sounds good!

Dockerfile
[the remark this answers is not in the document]
aminfa: I removed tini

docker-buildscript.sh
aminfa: I removed the build script. The created label is now set by the official build-push-action action

docker-buildscript.sh
aminfa: I removed the buildscript

Dockerfile
aminfa: I removed tini and added the usual suspects to `platforms` in the github actions ci.

.dockerignore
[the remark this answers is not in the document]
aminfa: Done

koraa:
> If we don’t need multiple Dockerfiles, keeping everything in the root directory is simpler and more conventional, as it aligns with common practices (tools and CI/CD systems typically expect the Dockerfile in the root). It also means we can run docker build . without needing to specify a path.

The point is that most people using rosenpass will be more interested in the rosenpass project rather than the docker container and right now there is a lot of stuff being added to the root repo which makes it harder to see anything.

You are right that having docker stuff no in root is less convenient for those users mostly interested in the docker container, but I think this is not the main consumer of the repo.

aminfa:
This PR should be ready now. It now adds a GitHub Actions workflow (`.github/workflows/docker.yaml`) to build and publish two Docker images: `ghcr.io/rosenpass/rosenpass` and `ghcr.io/rosenpass/rp`. 

The workflow runs when pushing to the main branch or when pushing a tag that matches `v*`, and also on pull requests (but without pushing the images). 

It uses `docker/metadata-action` to generate tags (`edge` for the main branch, short commit SHA, or semver if the push is a tagged release), then builds images for both amd64 and arm64. The multi-stage Dockerfile in the .docker directory produces two final stages: one for the standalone rosenpass binary and one for the integrated rp tool.

If I am not mistaken, the error's in the CI seem to be unrelated.

.docker/README.md
```diff
@@ -0,0 +1,236 @@
+# Rosenpass
```
koraa, line 1: Can you refer to this file from the main README?

koraa:
@aminfa Please rebase your PR

koraa:
@aminfa And thank you :)

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: .docker/Dockerfile
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          target: rosenpass
+          platforms: amd64,arm64
```
baentsch, line 93: fwiw, I'd agree that this is the cleanest/most "orthogonal" way to do things -- but if I'm not mistaken (I may be), at least one image is going to be built by emulation, i.e., the build is taking pretty long. An alternative approach would be to build on dedicated runners `aarch64` and `x64` images, respectively, and then merge them (their metadata) during push: That should be much more resource efficient/faster. The code for that is uglier, though :/ Or maybe a compromise: Only build (as a test for the logic) the native image during PR and both (platforms) only when merging/pushing to main?

.docker/README.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+Done! The two containers should now be connected through a wireguard VPN (Port 1000) with pre-shared keys exchanged by rosenpass (Port 9999).
+
+Now, test the connection by starting a shell inside the client container, and ping the server through the VPN:
+
+```bash
+# client
+docker exec -it rpclient bash
+apt update; apt install iputils-ping
```
baentsch, line 206: Why not include this util straight in the image? Seems helpful for any user.

.docker/README.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+It is now possible to set add these keys as pre-shared keys within a wireguard interface.
+
+```bash
+PREKEY=$(cat workdir/client-sharedkey)
+wg set <interface> peer <peer-public-key> preshared-key <(echo "$PREKEY")
+```
+
+## Usage - Combined with wireguard
```
baentsch, line 129: I like this complete usage documentation very much. Might it be sensible to pull this into a separate file USAGE.md that could then be uploaded to the (any) docker image repo as documentation for the users not interested in building stuff but just using it? And as a final suggestion, what about adding a script doing all of the below? It could serve two purposes: 1) Letting the unwary (or impatient) user run it all in "one quick go" (to convince "the unwary" that everything is easy) and 2) use in CI (to convince the development team that everything is in good working order).

aminfa: > I like this complete usage documentation very much. Might it be sensible to pull this into a separate file USAGE.md that could then be uploaded to the (any) docker image repo as documentation for the users not interested in building stuff but just using it?

Hm, this documentation ist mostly just a usage guide anyway. I can gladly rename it to `.docker/USAGE.md` if you want and restructure it to have the build section at the bottom. This way there is a higher focus on how to use the images.

aminfa: > And as a final suggestion, what about adding a script doing all of the below? It could serve two purposes: 1) Letting the unwary (or impatient) user run it all in "one quick go" (to convince "the unwary" that everything is easy)

I can see the appeal of a single script that simplifies everything for newer or impatient users. However, the examples in the current usage guide mainly demonstrate a niche examplary setup with two containers on the same system. This doesn’t translate into a script that would be broadly useful in practice.

A truly “all-in-one” solution, like what [PiVPN](https://github.com/pivpn/pivpn) provides, would require a lot more work to ensure correct configuration, security, and integration, and I feel that’s beyond the scope of this PR. If we ever introduce such a script (akin to PiVPN) in this repository, then we could certainly bundle it into the Docker images in a future PR. But for now, I’d rather keep this PR focused on bundling the two main rosenpass commands in docker images.

aminfa: > and 2) use in CI (to convince the development team that everything is in good working order).

hm, let me see if this is do-able

.docker/README.md
aminfa: Since it isn’t necessary for the container’s core functionality, adding it by default would bloat the image size, introduce an extra dependency which potentially increase the attack surface. 

For those who need it, installing iputils-ping is simple enough in an interactive session. Or it is fairly easy to create a new `Dockerfile` that extends our minimal images with utilities that they need. 

But if you want I can easily add the installation of `iputils-ping` to the Dockerfile.

.github/workflows/docker.yaml
aminfa: Good point :+1: 

I have added a strategy matrix to the workflow that allows building arm images natively on `ubuntu-latest-arm64` 
Let's see if it works

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,93 @@
+name: ci
```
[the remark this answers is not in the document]
aminfa: Good catch. I fixed it :+1:

.docker/README.md
aminfa: I have now added the `build-and-test-rp` job to the `docker.yaml` workflow. This performs an integration test by letting a rosenpass server and client instance exchange keys and comparing the resulting shared key files  for equality.

.github/workflows/docker.yaml
aminfa: The `runs-on: ubuntu-latest-arm64` doesn't seem to be picked up by a runner. It seems the runner may need to be activated in the organisational settings (@koraa): 

> To get started using Ubuntu 24.04, create an arm64 runner in your organization/enterprise, and select the “Ubuntu 24.04 by Arm Limited” partner image. 
> source: https://github.blog/changelog/2024-06-24-github-actions-ubuntu-24-04-image-now-available-for-arm64-runners/

If this option is not available, we can instead only run on the default x86 runner and only build on QEMU if CI-run is not a PR.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,198 @@
...
+
+jobs:
+  # --------------------------------
+  # 1. BUILD & TEST
+  # --------------------------------
+  build-and-test-rp:
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+    runs-on: ${{ matrix.arch == 'arm64' && 'ubuntu-latest-arm64' || 'ubuntu-latest' }}
```
baentsch, line 21: Is this "runs-on" designation the right name for "arm64"? According to [this](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners), the name IMO must read "ubuntu-24.04-arm" to actually run on `arch64` HW. Otherwise, this might be an emulation on `x64` (the CI runtime should be tell-tale).

.github/workflows/docker.yaml
baentsch: Oops -- sorry I see this only now -- see separate comment I just left: It might be as simple as changing the runner-name.

.github/workflows/docker.yaml
aminfa: Good point, let me try  it out

aminfa: No, doesn't seem to work. It may need to be enabled in the organisational settings

.docker/README.md
baentsch: > But if you want I can easily add the installation of iputils-ping to the Dockerfile.

No, you've got good arguments to keep it out. Also, the UID of the docker image is root, so user installation of this (and anything else :) is possible. Consider resolved.

.github/workflows/docker.yaml
aminfa: Nevermind, I missed a spot. It now uses the correct runner name for all jobs in the docker build workflow.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,198 @@
...
+            org.opencontainers.image.title=Rosenpass
+            org.opencontainers.image.description=Reference implementation of the protocol rosenpass protocol
+            org.opencontainers.image.vendor=Rosenpass e.V.
+            org.opencontainers.image.licenses=MIT OR Apache-2.0"
+            org.opencontainers.image.url=https://rosenpass.eu
+            org.opencontainers.image.documentation=https://rosenpass.eu/docs/
+            org.opencontainers.image.source=https://github.com/rosenpass/rosenpass
+      - name: Log in to registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+      - name: Set up QEMU
```
baentsch, line 185: Is QEMU still needed now that (only?) native runners are de/employed?

.github/workflows/docker.yaml
baentsch: And it now seems to complete in below 4mins as opposed to an hour before if I looked right: LGTM & the environment for saved compute cycles and developers for less wait CI time both say: Thanks @aminfa !

.github/workflows/docker.yaml
aminfa: No, I just missed removing this step in the second job. Thanks :)

.github/workflows/docker.yaml
aminfa: Yeah, it was a good idea. Thank you

Dockerfile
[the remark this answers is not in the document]
aparcar: @koraa comments on handling of SIGKILL? Does `rp` handle it wrong?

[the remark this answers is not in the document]
koraa: Is it really SIGKILL or is it SIGINT?

SIGKILL can not be caught, but it is quite possible that we are omitting proper SIGINT handling…signal handling in Rosenpass is generally a mess right now.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,196 @@
...
+
+on:
+  push:
+    branches:
+      - "main"
+    tags:
+      - "v*"
+  pull_request:
+    branches:
+      - "main"
```
aparcar, line 11: Please run this only on pull requests which actually touch the Dockerfile or docker.yaml. Else this burn a lot of precious CI

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,196 @@
...
+            type=edge,branch=main
+            type=sha,branch=main
+            type=semver,pattern={{version}}
+          labels: |
+            maintainer=Karolin Varner <karo@cupdev.net>, wucke13 <wucke13@gmail.com>
+            org.opencontainers.image.authors=Karolin Varner <karo@cupdev.net>, wucke13 <wucke13@gmail.com>
+            org.opencontainers.image.title=Rosenpass
+            org.opencontainers.image.description=The rp command-line integrates Rosenpass and WireGuard to help you create a VPN
+            org.opencontainers.image.vendor=Rosenpass e.V.
+            org.opencontainers.image.licenses=MIT OR Apache-2.0"
```
aparcar, line 135: Is here a stray quote?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,196 @@
...
+            type=edge,branch=main
+            type=sha,branch=main
+            type=semver,pattern={{version}}
+          labels: |
+            maintainer=Karolin Varner <karo@cupdev.net>, wucke13 <wucke13@gmail.com>
+            org.opencontainers.image.authors=Karolin Varner <karo@cupdev.net>, wucke13 <wucke13@gmail.com>
+            org.opencontainers.image.title=Rosenpass
+            org.opencontainers.image.description=Reference implementation of the protocol rosenpass protocol
+            org.opencontainers.image.vendor=Rosenpass e.V.
+            org.opencontainers.image.licenses=MIT OR Apache-2.0"
```
aparcar, line 179: Quote again

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,196 @@
...
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: .docker/Dockerfile
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          target: rp
+          platforms: linux/${{ matrix.arch }}
+  docker-image-rosenpass:
```
aparcar, line 153: Since we don't push this for PRs anyway and test it prior, why build it here again?

aparcar:
> The point is that most people using rosenpass will be more interested in the rosenpass project rather than the docker container and right now there is a lot of stuff being added to the root repo which makes it harder to see anything.

I'd add at least the Dockerfile to the root of the repository, it gives good visibility that we do support it and looking at new projects, I rather try them in a container rather than bare metal.

Secondly I think once we have proper container and broker API (currently experimental), we should make Rosenpass a thing on platforms that don't easily support Rust (i.e. OpenWrt). Instead the new package manager support running containers, a fine fit for Rosenpass.

.github/workflows/docker.yaml
aminfa: Yes, fixed it. Thanks

.github/workflows/docker.yaml
aminfa: Done.  I’ve added logic to run the Docker build only when the Dockerfile or docker.yaml changes, which definitely saves CI resources. 

**However**: I’m slightly concerned that changes to the source code (e.g., the Cargo workspace) could still break the build or produce a non-functional Docker image, yet we wouldn’t catch that if we skip the Docker build whenever the Dockerfile remains unchanged

.github/workflows/docker.yaml
aminfa: Good point. I add an `if` to the second job that disables building the same image again

.github/workflows/docker.yaml
aparcar: If we break Rosenpass the rest of the CI tests should find this, too, right?

aminfa: Sure. But in this unlikely scenario you might need to fix the Dockerfile too and the CI wouldn't tell you

baentsch: With the improvement to using native runners, the delta-resources the docker build requires is imo negligible compared to the benefit of an end-to-end user-ready integration-cum-test this PR now provides: This is a [good 3 & 4 minutes, in parallel, respectively](https://github.com/rosenpass/rosenpass/actions/runs/13525211906) -- compared to [66 mins just for (one of several such long-running) nix builds](https://github.com/rosenpass/rosenpass/actions/runs/13517812464/job/37784565146): The latter would arguably be CI runs that ought to garner attention for improving resource utilization and PR turnaround times.

aminfa:
Thanks, @aparcar and @baentsch, for taking the time to review this! I’ve incorporated all the requested changes, and CI is now passing. Regarding the file names and locations of the Dockerfile and its documentation, I see this as a matter of taste. I'm happy to adjust them based on the final consensus. Let me know if there's anything else you'd like to tweak!

Dockerfile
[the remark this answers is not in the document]
aminfa: Sorry I indeed meant SIGINT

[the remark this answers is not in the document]
aminfa: Also SIGTERM is not handled correctly. If I use `docker stop -t 100 rpclient` it stops the container after 100 seconds. Which means that the SIGTERM signal was ignored, and then after 100 seconds docker sends SIGKILL to terminate the process

koraa:
@aminfa Is there any reason to use `./.docker` over `./docker`?

aminfa:
> Is there any reason to use ./.docker over ./docker?

No there isn't. We can choose whatever filenames we want

koraa:
Then lets do ./docker and be done with it @aminfa :)

Any vetos @baentsch @aparcar @aminfa ?

koraa:
> > The point is that most people using rosenpass will be more interested in the rosenpass project rather than the docker container and right now there is a lot of stuff being added to the root repo which makes it harder to see anything.
> 
> I'd add at least the Dockerfile to the root of the repository, it gives good visibility that we do support it and looking at new projects, I rather try them in a container rather than bare metal.
> 
> Secondly I think once we have proper container and broker API (currently experimental), we should make Rosenpass a thing on platforms that don't easily support Rust (i.e. OpenWrt). Instead the new package manager support running containers, a fine fit for Rosenpass.

Lets merge now and clean up later!

aminfa:
Thanks for merging @koraa :)

But my apologies: the build is failing after merging into main because the workflow doesn’t have permissions to push to packages by default. I had previously tested it in our repository ([link to our test package](https://github.com/utilacy/rosenpass/pkgs/container/rp)), but I used hardcoded tokens with the necessary permissions, which isn't the case in the main workflow. I have pushed on update

PR #586, opened 2025-02-04
fix: Input dependent memory access in statistical constant time execu…

koraa:
…tion test

Supplying different memory locations to the memcmp function in the test is not constant time; this was an issue wit the test and not with memcmp itself.

The issue mainly showed up in the Release build where the correlation coefficient was in the ballpark of just below -0.01 with enough variance to sometimes but not usually fail the test. The precise reason for this is unknown but some optimization applied in the release build is most likely the culprit.

This should increase the stability of our CI which occasionally was flaky.


constant-time/src/memcmp.rs
```diff
@@ -44,20 +45,17 @@ mod tests {
     /// This test function will run an equal amount of comparisons on two different sets of parameters:
     /// - completely equal slices
     /// - completely unequal slices.
-    /// All comparisons are executed in a randomized order. The test will fail if one of the
```
phildremi, line 47: Was this deleted on purpose? The [remaining comment](https://github.com/rosenpass/rosenpass/blob/377f2f40d28327631890f211898520708a6507fd/constant-time/src/memcmp.rs#L48) doesn't seem to be self-sufficient.

koraa: Thanks for the catch

PR #604, opened 2025-02-10
[WIP] feature(rosenpass): integrate assert_tv and derandomize protocol.rs

aminfa:
Here is a WIP solution to integrate the [assert_tv](https://github.com/aminfa/assert_tv/tree/dev/v0.4) crate and derandomize the protocol.rs module.

This PR is mainly meant as a starting point to explore how  to make derandomization as non-disruptiv as possible and also not clutter the entire code-base with `#cfg[ ]` macros and feature flags.

The way derandomization would work is with the help of 3 macros:
- `#[test_vec(..)]`: proc macro that wraps a test and sets up a test_vec environment, which is stored and later accessed through thread-local storage, and finilizes it at the end. 
- `tv_const!(..)`: macro that defines a constant value in the test vector. 
- `tv_output!(..)`: macro that checks a result with the one found in the test vector.

Here is a `test_vec` test: [crypto_server_test_vectors.rs](https://github.com/utilacy/rosenpass/blob/dev/amin/derandomize/rosenpass/tests/crypto_server_test_vectors.rs) 

You may run this test with the following command

```
cargo test --package rosenpass --test crypto_server_test_vectors crypto_server_test_vector_1 --features tv -- --exact --ignored
```

Test vectors are frozen as files and checked into the repository. For example: [crypto_server_test_vector_1.json](https://github.com/utilacy/rosenpass/blob/dev/amin/derandomize/rosenpass/.test_vectors/crypto_server_test_vector_1.json)

You may run this test in `init` mode (by setting an env variable) to overwrite the values stores in this test vector file:

```
TEST_MODE=check cargo test --package rosenpass --test crypto_server_test_vectors crypto_server_test_vector_1 --features tv -- --exact --ignored
```

Let me know what you think about the proposal, what you like to see improved or changed, or if you have any questions.


aminfa:
I had noticed that the rosenpass CI uses `--all-features` option when testing, which would enable the `tv` feature flag that would compile in the entire test vector implementation. 

I have fixed this design error by removing the requirement to add a feature as a client of `assert_tv`. Now when testing the test vectors, you would need to enable the `assert_tv/enabled` feature flag instead. 

```
cargo test --package rosenpass --test crypto_server_test_vectors crypto_server_test_vector_1 --features assert_tv/enabled -- --exact
```

This let to a more streamlined integration elsewhere as well and many `#[cfg(..)]` macros also disappeared.

phildremi:
> I had noticed that the rosenpass CI uses `--all-features` option when testing, which would enable the `tv` feature flag that would compile in the entire test vector implementation, which would break existing working tests. (as false positives)
> 
> I have fixed this design error by removing the requirement to add a feature as a client of `assert_tv`. Now when testing the test vectors, you would need to enable the `assert_tv/enabled` feature flag instead.

I'm not familiar with `tv`, but if specific flags should be set in some scenarios and not doing it will lead to failures - maybe [CONTRIBUTING.md](https://github.com/rosenpass/rosenpass/blob/main/CONTRIBUTING.md) could be updated to explain this to new/potential contributors? It could link to the CI workflow as an authoritative reference if keeping both in sync is too much work (or the flags don't require much of an explanation).

PR #616, opened 2025-02-26
feat(docker): change write permission on docker build workflow to have write permission to packages

aminfa:
This is a small fix to the previous PR #582 . It now properly sets [the required permissions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions) to push the image. 

Apologies for missing this.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -10,6 +10,10 @@ on:
     branches:
       - "main"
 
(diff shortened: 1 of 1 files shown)


koraa:
@aminfa Thank you!

PR #619, opened 2025-03-05
Docker CI multi platform fix and cleanup

aparcar:
* Multi platform support is broken right now, fix it
* Cleanup use of unnecessary role

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -129,15 +129,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
       - name: Docker meta
         id: meta
+
         uses: docker/metadata-action@v5
         with:
-          images: ghcr.io/rosenpass/rp
-          tags: |
-            type=edge,branch=main
-            type=sha,branch=main
-            type=semver,pattern={{version}}
+          images: ghcr.io/${{ github.actor }}/rp
           labels: |
             maintainer=Karolin Varner <karo@cupdev.net>, wucke13 <wucke13@gmail.com>
             org.opencontainers.image.authors=Karolin Varner <karo@cupdev.net>, wucke13 <wucke13@gmail.com>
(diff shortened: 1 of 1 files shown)


aparcar:
@aminfa please have a look I'd be curious on your opinion

aminfa:
> please have a look I'd be curious on your opinion

Sure! From looking through the code, this looks good to me (though I haven't run it).

> [Commit e40054b](https://github.com/rosenpass/rosenpass/pull/619/commits/e40054bf285546acfe3a2b1f41edaf4dd47cbc77): Don't pull in an external action but rely on GitHubs native way to detect file changes.

I didn't know about this feature. Yes this is much cleaner!

> [Commit e40054b](https://github.com/rosenpass/rosenpass/pull/619/commits/e40054bf285546acfe3a2b1f41edaf4dd47cbc77): Also fix a logic flaw where a PR would try to push an image (but never succeed due to missing secrets).

I am not sure what this is referencing, to be honest.

aminfa:
> [commit 7f66ccd8](https://github.com/rosenpass/rosenpass/pull/619/commits/7f66ccd8e892fa6ab0990f45d191cde7d7f375fd) The previous approach would use a CI matrix to build containers for multiple platforms (arm64, amd64). Since both use the same label/tag, the job running second will overwrite the prior.

Just as a heads-up, this seems to undo some of the changes that were [previously discussed](https://github.com/rosenpass/rosenpass/pull/582#discussion_r1967739098) in the previous PR.

Also, I could be mistaken but I don't this is happening. When you take a look at the rosenpass package [https://github.com/rosenpass/rosenpass/pkgs/container/rosenpass] you can see two images. The image for the amd64-platform however seems to be not be recognized as such. Digging deeper this could relate to attestation done by the official docker action and its incompatibility with ghcr.io. (Sources: [1](https://github.com/docker/build-push-action/issues/820#issuecomment-1445131521), [2](https://github.com/orgs/community/discussions/45969#discussioncomment-4852744))

Nevertheless, I would say let's merge this PR and see if it indeed fixed the images on ghcr.io. With github-workflows it is sometimes trial and error.

aparcar:
> Just as a heads-up, this seems to undo some of the changes that were https://github.com/rosenpass/rosenpass/pull/582#discussion_r1967739098 in the previous PR.

Fair, I reworked it. 

> Also, I could be mistaken but I don't think this is happening. When you take a look at the [rosenpass package](https://github.com/rosenpass/rosenpass/pkgs/container/rosenpass) you can see two images. The image for the amd64-platform however seems to be not be recognized as such. Digging deeper this could relate to attestation done by the official docker action and its incompatibility with ghcr.io. (Sources: https://github.com/docker/build-push-action/issues/820#issuecomment-1445131521, [2](https://github.com/orgs/community/discussions/45969#discussioncomment-4852744))

From my understanding a label can contain multiple architectures and only one manifest can claim a label. If two single manifest try to take the same label, the latter wins (i.e. overwrites). 

> Nevertheless, I would say let's merge this PR and see if it indeed fixed the images on ghcr.io. With github-workflows it is sometimes trial and error.

I'll merge this since it appears to be [working](https://github.com/aparcar/rosenpass/pkgs/container/rosenpass/371963791?tag=main). I recommend always testing CI things on one's own repository CI first, not merging things on the main repository/branch until it works :)

PR #630, opened 2025-04-03
ci(supply-chain+dependabot): Automatically create exemptions for cargo-crev for dependa-bot PRs

DavidNiehues:
This PR adds a new CI workflow that only triggers for PRs by dependabot. When triggered, it regenerates the exemptions for cargo-crev so that the cargo-crev does not fail for dependabot PR and these can be merged.

Since the CI job only triggers for dependabot, this has to be tested on the main branch.

### DIFF
--- a/.github/workflows/cargo-crev-exemptions-dependabot.yml
+++ b/.github/workflows/cargo-crev-exemptions-dependabot.yml
@@ -1,0 +1,58 @@
+name: Dependabot Vet Exemptions
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "Cargo.toml"
+      - "Cargo.lock"
+
+jobs:
+  dependabot-cargo-crev-exceptions:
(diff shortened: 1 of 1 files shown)


aparcar:
Dear @DavidNiehues , thanks again for your contribution. When I saw the Bots PRs failing in the past I came up with a similar idea, however abandoned it since. From my understanding this generates a "blank check" for any package changes.

Could it be a more secure approach to specifically trust entities and rely on their work actually checking the code? For example, running `cargo vet trust --all epage` will allow any `clap` packages maintained and released by @epage. This would require a manual review of authors and organizations to trust once but then allow regular Bot PRs for package bumps.

koraa:
@aparcar Can you add a separate issue for this? I think we need to work on improving our practices regarding supply chain security slowly; for now, being able to update again is more crucial

aparcar:
Sure will do, please feel free to proceed with this

koraa:
Merged through #635

PR #667, opened 2025-06-26
Use serde for JSON-encoding benchmark data

keks:
This PR cleans up the way we serialize the JSON in the trace-based protocol benchmarks. We used to do manual serialization based on writing strings, but now we use serde.

### DIFF
--- a/rosenpass/benches/trace_handshake.rs
+++ b/rosenpass/benches/trace_handshake.rs
@@ -1,6 +1,9 @@
-use std::io::{self, Write};
-use std::time::{Duration, Instant};
-use std::{collections::HashMap, hint::black_box, ops::DerefMut};
+use std::{
+    collections::HashMap,
+    hint::black_box,
+    ops::DerefMut,
+    time::{Duration, Instant},
+};
 
 use anyhow::Result;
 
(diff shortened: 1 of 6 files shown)


rosenpass/benches/trace_handshake.rs
```diff
@@ -362,32 +344,33 @@ impl AggregateStat<Duration> {
...
-            name = label,                          // Benchmark name
-            value = self.mean_duration.as_nanos(), // Mean duration in nanoseconds
-            range = self.sd_duration.as_nanos(),   // Standard deviation in nanoseconds
-            sample_size = self.sample_size,        // Number of samples
-            os = std::env::consts::OS,             // Operating system
-            arch = std::env::consts::ARCH,         // CPU architecture
-            run_time = run_time_group(label),      // Run time group category (long, medium, etc.)
-            protocol_version = protocol_version // Overall protocol_version (e.g., protocol version)
-        )
+impl<'a> serde::Serialize for JsonAggregateStat<'a, Duration> {
```
franziskuskiefer, line 355: Can't you derive this implementation? `AggregateStat` and `T` would need to implement serialize then as well. So I don't know if it's really better.

franziskuskiefer (commented):
lgtm with one question

PR #678, opened 2025-07-30
fix Ctrl+C for rp

xavieryin:
Fix existing problems:

1. cleanup procedures fires up simutaneously => now LIFO
2. AppServer event_loop keeps running => terminates w/ AppServerTest termination_handler
3. possible racing cleanup => resolved using atomic flag


xavieryin:
It's fine that this is not accepted. I did notice that I reused existing test-only code. At least we've worked on a true issue.
I'll test and report under #685 today

PR #699, opened 2025-08-30
feat: add preliminary miri support

wucke13:
- Adds a devShell with Miri
- Marks some of the tests which Miri cannot execute as ignored for Miri


wucke13:
@koraa thank you!

Issue #79, opened 2023-05-24
Network manager support: Open vSwitch


koraa:
https://wiki.archlinux.org/title/Open_vSwitch

algora-pbc:
💎 **$150** bounty created by @koraa
👉 To claim this bounty, [submit your pull request on Algora](https://console.algora.io/bounties/cllxcxqj60012ju0fc7r0h4nr)
📝 Before proceeding, please make sure you can [**receive payouts**](https://docs.algora.io/bounties/payments#country-support) in your country
💵 Payment arrives in your account 2-5 days after the bounty is rewarded
💯 You keep 100% of the bounty award
🙏 Thank you for contributing to rosenpass/rosenpass!

jacanidjar:
@koraa This bounty is available, can I start it?

krapcys1-maker:
Hello its  available to do ?

yuliuyi717-ux:
/attempt #79

I can take this bounty.

Implementation plan:
1) Confirm current integration surface for Open vSwitch in Rosenpass and map expected operator workflow (bring-up, key exchange, route handling, teardown).
2) Add an Open vSwitch-focused integration path (service/config templates + docs) without breaking existing NetworkManager/systemd-networkd flows.
3) Implement robust config validation and clear error handling for common OVS misconfiguration cases.
4) Add tests and a reproducible example setup so maintainers can verify behavior quickly.

I will submit a focused PR with demo/proof according to the bounty checklist.

Harry-jain:
/attempt #79

Hey, I’d like to take this up.

From what I understand, the goal is to add Open vSwitch support as an alternative network backend in Rosenpass, similar to how NetworkManager/systemd-networkd are currently handled.

My rough plan:

- First I’ll go through the existing network integration code to understand where interfaces are created and managed (especially WireGuard setup)
- Then I’ll try adding a basic OVS integration using `ovs-vsctl` (create bridge if needed + attach wg interface)
- Extend config to allow something like selecting "ovs" backend with a bridge name
- Hook this into the lifecycle so that on startup it attaches the interface and on shutdown cleans it up (might take a couple tries to get this right)
- Add some basic checks (like if ovs is not installed or bridge doesn’t exist)
- Finally, I’ll add a small reproducible example + docs so it’s easy to test

I’ll also include a demo video as required once everything is working.

Let me know if there are any preferences around how the config should look or how deep the integration is expected to go.

yemsy26:
[...]

PR #717, opened 2026-03-10
feat: add Open vSwitch network manager support (#79)

davidequarracino:
Hi! I've implemented the Open vSwitch support requested in #79.

I followed the existing `systemd-networkd` pattern to keep the integration consistent with the project's current structure.

Summary:
- Added `config-examples/openvswitch/setup.sh`.
- Automated bridge/port management via `ovs-vsctl`.
- Kept the standard key generation flow.

Attached a quick demo of the script execution. Let me know if you need any changes.

/claim #79


davidequarracino:
Quick demo for the OVS script. Correctly hitting the system paths and parsing the bridge name.

https://github.com/user-attachments/assets/c83f6609-9a30-4c68-9479-a7316d9a3956

lyndon050516:
One functional issue with the OVS setup script: it adds `wg0` as an OVS `type=internal` interface:

```sh
ovs-vsctl --may-exist add-port "${BRIDGE}" "${INTERFACE}" -- set interface "${INTERFACE}" type=internal
```

That creates an OVS internal netdev named `wg0`, not a Linux WireGuard interface. The rest of the Rosenpass/WireGuard path expects `wg0` to be `type wireguard` so that `wg set wg0 peer ... preshared-key ...` can succeed. A user following this script can end up with an interface named `wg0` on the bridge, but it will not be usable by WireGuard/Rosenpass as the tunnel device.

For this integration I think the script needs to create/configure the WireGuard device first, then add that existing WireGuard netdev as the OVS port, without `type=internal`.

ilka-schulz:
This PR is decoherent in several regards, probably AI-generated.

Issue #82, opened 2023-05-24
Network manager support: ConnMan


koraa:
https://wiki.archlinux.org/title/ConnMan

algora-pbc:
## 💎 $150 bounty [• Rosenpass](https://algora.io/rosenpass)
### Steps to solve:
1. **Start working**: Comment `/attempt #82` with your implementation plan
2. **Submit work**: Create a pull request including `/claim #82` in the PR body to claim the bounty
3. **Receive payment**: 100% of the bounty is received 2-5 days post-reward. [Make sure you are eligible for payouts](https://algora.io/docs/payments#supported-countries-regions)

### ❗ Important guidelines:
- To claim a bounty, you need to **provide a short demo video** of your changes in your pull request
- If anything is unclear, **ask for clarification** before starting as this will help avoid potential rework
- Low quality AI PRs will not receive review and will be closed
- Do not ask to be assigned unless you've contributed before

Thank you for contributing to rosenpass/rosenpass!

| Attempt | Started (UTC) | Solution | Actions |
| --- | --- | --- | --- |
| 🟢 @a4to | Mar 02, 2026, 02:42:23 AM | #714 | [Reward](https://algora.io/claims/GvfyQk2CFQUjCFZa) |
| 🟢 @algora-pbc | Mar 02, 2026, 02:42:24 AM | WIP |  |
| 🟢 @pqcfox | Jan 07, 2024, 09:28:52 PM | WIP |  |
| 🔴 @aazam-gh | Sep 09, 2023, 09:17:50 AM | WIP |  |
| 🟢 @davidequarracino | Mar 10, 2026, 02:10:26 AM | #718 | [Reward](https://algora.io/claims/3RERdmUoNAHSycFy) |
| 🟢 @blessuselessk | Mar 21, 2026, 04:52:40 AM | #722 | [Reward](https://algora.io/claims/4KMkqQkNpWn1d7rd) |
| 🟢 @zzinterest | Mar 29, 2026, 12:32:23 PM | #724 | [Reward](https://algora.io/claims/HgYi1G6235x5VFiU) |

aazam-gh:
is rosenpass available for ubuntu?

emilengler:
> is rosenpass available for ubuntu? 

You have to compile it for yourself.

aazam-gh:
Would I need an arch Linux distro to solve this issue?

emilengler:
> Would I need an arch Linux distro to solve this issue?

[...]

PR #718, opened 2026-03-10
feat: minimalist ConnMan integration support (#82)

davidequarracino:
Hi! I saw the ongoing discussion on #82. 

Instead of adding a whole new infrastructure, I’ve put together a single, lightweight setup script. I think it’s better to keep things simple: it just handles the keys and generates a basic ConnMan provisioning file, following the same pattern already used in the repo for other network managers.

It’s less than 30 lines of code—easy to review, easy to maintain, and get the job done without any bloat.

Let me know if this approach works for you!

/claim #82

### DIFF
--- a/config-examples/connman/setup.sh
+++ b/config-examples/connman/setup.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+# Rosenpass + ConnMan Integration (Minimalist)
+set -e
+
+INTERFACE="${1:-wg0}"
+ROLE="${2:-server}"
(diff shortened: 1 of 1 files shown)


davidequarracino:
Quick demo attached. Scripts works as intended, correctly hitting the system paths and parsing arguments.

https://github.com/user-attachments/assets/0b99728b-7f0f-41d4-bc2e-88e5fbeb7c5c

ilka-schulz:
As far as I understand ConnMan, you need to define a ConnMan VPN driver to support an additional VPN technology. This PR seems unrelated to me.

Issue #80, opened 2023-05-24
Network manager support: netctl


koraa:
https://wiki.archlinux.org/title/Netctl

algora-pbc:
## 💎 $150 bounty [• Rosenpass](https://algora.io/rosenpass)
### Steps to solve:
1. **Start working**: Comment `/attempt #80` with your implementation plan
2. **Submit work**: Create a pull request including `/claim #80` in the PR body to claim the bounty
3. **Receive payment**: 100% of the bounty is received 2-5 days post-reward. [Make sure you are eligible for payouts](https://algora.io/docs/payments#supported-countries-regions)

### ❗ Important guidelines:
- To claim a bounty, you need to **provide a short demo video** of your changes in your pull request
- If anything is unclear, **ask for clarification** before starting as this will help avoid potential rework
- Low quality AI PRs will not receive review and will be closed
- Do not ask to be assigned unless you've contributed before

Thank you for contributing to rosenpass/rosenpass!

| Attempt | Started (UTC) | Solution | Actions |
| --- | --- | --- | --- |
| 🟢 @MahmoudAdelbghany | Mar 08, 2026, 05:52:05 AM | #715 | [Reward](https://algora.io/claims/h4s6woyfNCLcGvk1) |
| 🟢 @algora-pbc | Mar 08, 2026, 05:46:44 AM | WIP |  |
| 🟢 @Hrushi20 | Oct 19, 2023, 11:55:07 AM | WIP |  |
| 🟢 @5m2wse50 | Mar 19, 2026, 11:49:53 PM | #720 | [Reward](https://algora.io/claims/JpPoHpGxXzwnFt5P) |
| 🟢 @zzinterest | Mar 29, 2026, 12:47:55 PM | #726 | [Reward](https://algora.io/claims/4juysA5iPRhTKK4Z) |

Hrushi20:
Hey @koraa, can you give me a brief description of this issue? I've read through netctl.

koraa:
> Hey @koraa, can you give me a brief description of this issue? I've read through netctl.

When I add a wifi network to my computer I can just create a new config file with netctl

```
Interface=wlp3s0
Connection=wireless
Security=wpa
…
```

It should be possible to create and manage simple rosenpass/wireguard setups (using `rp`) with netctl in a similar way.

[...]

PR #720, opened 2026-03-19
feat(broker): add netctl support for Arch Linux

5m2wse50:
This PR implements a `NetctlBroker` to support key distribution on Arch Linux systems using `netctl`.

### Changes
- Added `NetctlBroker` implementation in `wireguard-broker`.
- Added `netctl` configuration option to `rosenpass.toml`.
- Updated CLI to utilize the new broker when configured.

Tested locally by verifying profile updates and restart commands.

/claim #80


ilka-schulz:
As far as I understand `netctl`, you need to define a connection script with `<connection_type>_up`, etc. for a new `Connection` type. Instead, you are reading `/etc/netctl` and look for the literal line `Connection=wireguard` which is not appropriate.

PR #723, opened 2026-03-25
feat: add Open vSwitch (OVS) network integration backend

Harry-jain:
/claim #79

## Summary
This PR adds Open vSwitch (OVS) support as a network backend for Rosenpass.

## Features
- OVS bridge integration for WireGuard interfaces
- Automatic attach/detach of interfaces to OVS bridges
- Configurable backend via config file
- Robust validation and error handling

## Implementation
- Added OVS integration module using ovs-vsctl
- Extended config schema to support OVS backend
- Integrated lifecycle hooks for setup and teardown

## Demo
Attached video demonstrates:
- OVS bridge creation
- Rosenpass startup
- WireGuard interface attached to OVS
- Functional networking

## Notes
This implementation does not affect existing NetworkManager or systemd-networkd flows.

### DIFF
--- a/demo-ovs.sh
+++ b/demo-ovs.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+set -e
+
+echo "Starting OVS Demo with Rosenpass"
+
+# 1. Create OVS bridge manually (or rely on rosenpass auto-create)
+# We will rely on Rosenpass to auto-create it, but let's clear it first if it exists
+ovs-vsctl --if-exists del-br br0 || true
+ovs-vsctl add-br br0
+
+# Generate a config file with OVS enabled
+cat <<EOF > config.toml
+public_key = "rp-public-key"
+secret_key = "rp-secret-key"
+listen = ["0.0.0.0:9999"]
+verbosity = "Verbose"
+
+[network]
+backend = "ovs"
+bridge = "br0"
+interface = "wg0"
+
+[[peers]]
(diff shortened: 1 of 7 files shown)


Harry-jain:
@copilot apply changes based on the comments in [this thread](https://github.com/rosenpass/rosenpass/pull/723#pullrequestreview-4009931364)

lyndon050516:
One functional gap in the demo/docs: the example peer is configured with `key_out = "key-out.txt"`, but no `device`/`peer` fields. With that config Rosenpass writes the derived key to a file; it does not build the native WireGuard broker that calls `wg set <interface> peer <peer> preshared-key ...`.

So the OVS backend can create/attach `wg0`, but the documented demo still does not prove that Rosenpass is delivering PSKs into the WireGuard peer on that OVS-attached interface. A user following the docs would see `wg0` and `br0`, while `wg show wg0 preshared-keys` would remain unset unless they separately configured a peer and broker path.

I would suggest changing the demo/docs to use the normal WireGuard peer fields (`device = "wg0"`, `peer = "<WG public key>"`) and adding a verification step that checks `wg show wg0 preshared-keys` after exchange. That would exercise the actual bounty path: OVS port membership plus Rosenpass-managed WireGuard PSK rotation.

ilka-schulz:
This PR does not solve the problem, probably AI-generated.

PR #760, opened 2026-06-05
fix syntax error in GitHub Workflow doc-upload.yml introduced in 2ac84c718c43c4a5512798da39eb5a34f64975

ilka-schulz:


ilka-schulz:
The fixed test is only triggered on `main` branch. This is too much work for such a simple fix, so I will just merge it and check if it works.

PR #767, opened 2026-06-05
switch to using hash for immutability

loganaden:


ilka-schulz:
Ah, I understand now what you meant earlier. I completely understood you wrong. I will look into it.

ilka-schulz:
Getting back to your PR as promised: I think that purely switching from version numbers to hashes has the following impact:
1. We will get a very specific version of the package.
2. We will not automatically get patch updates.
3. We would not get Depedabot alerts anymore:
> Dependabot only creates alerts for vulnerable actions that use semantic versioning and will not create alerts for actions pinned to SHA values.

https://docs.github.com/en/enterprise-cloud@latest/actions/reference/security/secure-use

I think (2) poses a vulnerability which we have to mitigate somehow. Maybe a very frequently running Dependabot which upgrades the hash but that would defy the whole purpose as we do not have the capacity to manually check that every time. We would be specifing hashes blindly.

Also, (3) would make known vulnerabilities invisible which we would need to fix, too.

I would prefer not to merge this PR for now. Maybe you have a different view angle on this?

loganaden:
I understand. It's better to hold on.

ilka-schulz:
I will close this PR for now. Maybe if we find a good solution, we can come back to this again.

PR #794, opened 2026-06-08
upgrade libcrux* dependencies to latest versions

ilka-schulz:
This PR should probably never be merged. I went through their repo and we should probably drop `libcrux`. I will open a new PR (#795) where I remove `libcrux` from `rosenpass` and share the details of my findings.


ilka-schulz:
closed in favor of #795

Issue #788, opened 2026-06-07
clippy [EPIC]


ilka-schulz:
We should give more attention to [clippy](https://doc.rust-lang.org/stable/clippy/usage.html).

PR #810, opened 2026-06-09
upgrade dependency `rtnetlink` and fix regressions

ilka-schulz:
deprecates #805


rp/src/exchange.rs
ilka-schulz, line 117: Never mind that clippy warning, we have an issue for that: #788

Issue #843, opened 2026-06-15
go through cargo-vet exemptions and fix them


ilka-schulz:
We have lot of exemptions and I am sure that not every of those has been reviewed by a human. However, writing all those audits would be a huge amount of work...

also see #831

ilka-schulz:
As I understand it now, this is not really possible. We can not review every patch release, that would be like 20 reviews per week.

ilka-schulz:
some more thoughts on this here: https://github.com/rosenpass/rosenpass/issues/831#issuecomment-5238106131

PR #845, opened 2026-06-15
resolve #831: no longer auto-generate vet exemptions (require human review instead)

### LINKED ISSUE #831: decide on SOPs when dealing with `cargo vet` exemptions
We should keep a human in the loop for regenerating cargo-vet exemptions:

> There are several things that cargo vet can do for you automatically but we choose to make manual just to keep a human in the loop of those decisions. Some of these might one day become automatic if we agree they're boring/reliable enough.

https://mozilla.github.io/cargo-vet/commands.html

ilka-schulz:
closes #831 

alternative solution: #833

### DIFF
--- a/.github/workflows/supply-chain.yml
+++ b/.github/workflows/supply-chain.yml
@@ -41,15 +41,17 @@ jobs:
       - name: Generate cargo-supply-chain report about crates
         run: cargo supply-chain crates
     # The setup for cargo-vet follows the recommendations in the cargo-vet documentation: https://mozilla.github.io/cargo-vet/configuring-ci.html
+
+  # runs cargo vet in order to check for unvetted dependencies
+  # documentation: https://mozilla.github.io/cargo-vet/index.html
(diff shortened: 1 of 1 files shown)


ilka-schulz:
@koraa What do you think? I kept the old job in the file but commented it out. The rationale for this PR is explained in the linked issue #831.

I also opened #843 but I think we do not really have the capacity to do that now.

There is an alternative PR for the same issue by @leemeo3 in #833 which keeps the functionality to use `--regenerate-exemptions` at the end of a commit message to let GH Workflow create a commit which regenerates exemptions. We can alternatively go with that PR (maybe requires some cleanup from what I can see). However, I went through the repo and there is no commit which is an ancenstor of `main` which has this `--regenerate-exemptions` in its commit message.

ilka-schulz:
closing this unmerged because we probably want to keep this behavior for now. I marked the corresponding issue #831 as blocked for now.

PR #897, opened 2026-07-04
Group non-critical dependabot updates into weekly PRs

### LINKED ISSUE #896: dependabot should not create a PR for each non-critical patch upgrade but make one PR per week with non-critical patch upgrades
dependabot should not create a PR for each non-critical patch upgrade but make one PR per week with non-critical patch upgrades. Otherwise, we get such annoying PRs like:
- #893 
- #895

We could restrict dependabot to apply security critical upgrades and then we could make a `cargo update` once per week.

`cargo deny` already runs once per week, so that could be an anchor point.

zain-asif-dev:
Fixes #896.

Both the cargo and github-actions dependabot configs were on a daily schedule with no grouping, so every minor or patch bump gets its own PR (e.g. #893, #895). That's a lot of near-identical noise for updates that don't need individual review.

This switches both ecosystems to a weekly schedule and adds a `non-critical` group for minor and patch updates so they get bundled into one PR per week. Major version bumps aren't in the group, so they still show up as their own PR and get separate attention.

Tested by loading the updated `.github/dependabot.yml` with a YAML parser to confirm it's still valid.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,11 +3,21 @@ updates:
   - package-ecosystem: "cargo"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+    groups:
+      non-critical:
+        update-types:
+          - "minor"
+          - "patch"
     # open-pull-requests-limit: 0 # temporarily disable dependabot
 
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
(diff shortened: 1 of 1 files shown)


ilka-schulz:
I do not see how this PR solves the problem described in the corresponding issue. We would still get (almost) the same PRs (which means also the same amount of PRs) from dependabot but they would all appear weekly. I do not think that this would really decrease the workload.

PR #919, opened 2026-07-28
Add check for empty received message

loganaden:


ilka-schulz:
@loganaden Thank you for that find! I will probably not have time this week to merge it because I have some other urgent task to do but will come back to it next week.

loganaden:
@ilka-schulz Thank you for the review.

ilka-schulz:
I figure that the function does not need to be constant-time, so that obstacle is removed.

ilka-schulz:
I also checked that the change does not cause any unintended side-effects, so that obstacle is also removed.

ilka-schulz:
It might be a good idea to throw `RosenpassError::BufferSizeMismatch` instead as it is done by the rest of the code. @koraa What do you think?

loganaden:
I can update the patch.

ilka-schulz:
@loganaden Can you please use the exact same error message as here? https://github.com/rosenpass/rosenpass/blob/c6cafa8101c564ac550f0ed97ffe2eb0d227ce6a/rosenpass/src/protocol/protocol.rs#L2170

I would merge the PR afterwards.

loganaden:
@ilka-schulz is this OK ?

ilka-schulz (approved):
Looks good. I will merge it as soon as the CI has finished again.

PR #959, opened 2026-08-10
fix(nix): provide man for manual page tests

### LINKED ISSUE #840: Test runs fail on Debian 13 Minimal
Line 57 of `rosenpass/tests/main-fn-generates-manpages.rs` calls
`let res: std::process::Output = std::process::Command::new("man").arg(path).output()?;`
which makes the test (and by extension cargo test for the entire repo) fail if "man" is not installed on the system.
Probably a non-issue considering that it only affects development and having a system without man installed is quite unlikely.

mgalore:
## Summary

- add `man-db` to the default and full Nix development shells
- include `man-db` in the Rosenpass package test environment, where generated-manpage integration tests execute `man`

Fixes #840.

## Verification

- `git diff --check`
- `cargo metadata --no-deps --format-version=1`
- verified the Nix dependency declarations place `man-db` in both development shells and `nativeBuildInputs` for package tests

`RUST_MIN_STACK=8388608 cargo test -p rosenpass --test main-fn-generates-manpages --all-features` could not complete on this worker because the host lacks `libclang` (required by `oqs-sys` bindgen); the project Nix package already provides `rustPlatform.bindgenHook`, and Nix is unavailable on this worker for the corresponding Nix evaluation.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -145,6 +145,7 @@
                 rustfmt
                 prettier
                 nushell # for the .ci/gen-workflow-files.nu script
+                man-db # renders generated manual pages in Rust integration tests
                 proverif-patched
               ];
             };
@@ -161,6 +162,7 @@
                 rustfmt
                 prettier
                 nushell # for the .ci/gen-workflow-files.nu script
+                man-db # renders generated manual pages in Rust integration tests
                 proverif-patched
                 pkgs.cargo-llvm-cov
                 pkgs.grcov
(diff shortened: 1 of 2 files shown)


SFeroudj-iteratec:
Hi @mgalore, thanks for the contribution.
You're adding the `man-db` package where the `mandoc` package already provides `man`.
I don't see an advantage to providing `man` twice. Can you explain your reasoning?

mgalore:
You are right—`mandoc` already provides the `man` command needed by the test, so `man-db` is redundant. I do not have a valid reason to add it. Thank you for catching this; I am withdrawing this PR rather than keeping an unnecessary dependency change open.

mgalore:
Closing as withdrawn: the proposed dependency is redundant with `mandoc`.

PR #1017, opened 2026-08-27
fix `<` operator to `<=` operator in cookies doc test

ilka-schulz:
The previous assertion statement could lead to rare test failures, e.g. in https://github.com/rosenpass/rosenpass/actions/runs/33050859863/job/98445824160

The test was introduced in 46e855b266a45113207a732f622c191bc46f66d5.


koraa (approved):
Good catch; thank you!

PR #1041, opened 2026-08-31
fix: Ensure that the experiment_api feature compiles; ensure that doctests enable internal API.

### LINKED ISSUE #1025: Bug (development): Doctests failing
The doctests are failing unless the `expose_internal_modules` feature is enabled.

Fails:

```cargo test -p rosenpass --doc```

Works:

```
cargo test -p rosenpass --doc --features expose_internal_modules
```

We should make sure that:

- Doctests work (without features enabled), either by:
    - Disabling doctests which require the expose_internal_modules feature
    - Force-enabling the expose_internal_modules feature
- Add a nix flake check, which ensures this stays the case

@ilka-schulz

### PR DESCRIPTION
**Pull request by koraa**
The command

```
cargo check --workspace --features experiment_api --all-targets
```

would fail, because some of the broker binaries require the tokio dependencies and some require access to the internal API. Use of the internal API is now enabled for these features, and experiment_api pulls in the tokio dependency.

We also added a regression test to make sure the experiment_api feature compiles on its own.

---
fix: Doctests require expose_internal_modules feature

Also explicitly builds with just expose_internal_modules enabled to
ensure all the tests build in this configuration.

Fixes: #1025

### CODE REVIEW

File: rosenpass/src/api/boilerplate/message_trait.rs [hunk 1]
```diff
@@ -52,7 +52,8 @@ where
     ///
     /// # Examples
     ///
-    /// ```
+    #[cfg_attr(feature = "expose_internal_modules", doc = "```")]
+    #[cfg_attr(not(feature = "expose_internal_modules"), doc = "```ignore")]
```
ilka-schulz, line 56: This looks like a decent fix. I already contemplated this problem, too, and I would have implemented a fix the same way.

One thing: is it possible to stuff these two lines into one short macro? It would be less code repetition, less error-prone and I think that this one layer of abstraction (the additional macro) would probably not hurt.

rosenpass/src/internal/cipher_traits/primitives/kem.rs
```diff
@@ -16,7 +16,8 @@
 //! encapsulate a symmetric key and Alice to decapsulate it would look as follows.
 //! In the example, we are using Kyber512, but any KEM that correctly implements the [Kem]
 //! trait could be used as well.
-//!```rust
+#![cfg_attr(feature = "expose_internal_modules", doc = "```rust")]
+#![cfg_attr(not(feature = "expose_internal_modules"), doc = "```ignore")]
```
ilka-schulz: Can you please use `rust,ignore` instead of `rust`? (applies to other parts of this PR as well)

ilka-schulz (changes requested):
Nice fix! Take a look at the two comments I made about potential minor changes which could fix the PR up.

rosenpass/src/internal/cipher_traits/primitives/kem.rs
koraa: Can you apply this yourself?

rosenpass/src/api/boilerplate/message_trait.rs
koraa: Not unless we want to write an internal derive crate; the way we are doing it seems to be the state of the art.

https://github.com/rust-lang/rust/issues/133886

ilka-schulz (approved):

PR #1045, opened 2026-08-31
fix #620: rp script no longer fails in environments where `git` is not installed

### LINKED ISSUE #620: rp fails with error message if Git is not installed
Tested on v0.2.2 under Debian:

```sh
vboxuser@debian:~$ rp help
WARNING: /usr/local/bin/rp: line 50: git: command not found
Usage: /usr/local/bin/rp [explain] [verbose] genkey|pubkey|exchange [ARGS]...
```

We should
- [ ] check if the new Rust-rp also fails under that condition

ilka-schulz:
- closes #620

### DIFF
--- a/rp
+++ b/rp
@@ -43,15 +43,21 @@ dbg() {
   echo >&2 "$@"
 }
 
-
+# check whether the `${scriptdir}` is a git repository
+# stdout output:
+# - if it is a git repository: root path of the repo
+# - if it is not a git repo: empty output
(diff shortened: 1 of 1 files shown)


koraa (approved):
Can we add a WARNING if binaries from git are used?

ilka-schulz:
> Can we add a WARNING if binaries from git are used?

planned in #1047

Issue #957, opened 2026-08-10
fix clippy warnings


ilka-schulz:
has been started in #934, but there are still some warnings, e.g. in test configuration

If some warning is hard to fix, we should at least have an open issue and reference that issue in the code.

PR #1050, opened 2026-09-01
Fix/957 clippy warnings

SFeroudj-iteratec:
Issue [#957](https://github.com/rosenpass/rosenpass/issues/957)

Addresses a lot of the warnings when running `cargo clippy --all-features --all-targets`.

Not done:
* There are a lot of warnings for `missing_docs_in_private_items` which I don't feel qualified to address right now
* There is one warning for `missing_safety_doc` which, again, I don't feel qualified to address right now

### DIFF
--- a/rosenpass/src/internal/cipher_traits/primitives/aead.rs
+++ b/rosenpass/src/internal/cipher_traits/primitives/aead.rs
@@ -141,7 +141,7 @@ pub trait AeadWithNonceInCiphertext<
         let (nonce, rest) = ciphertext.split_at(NONCE_LEN);
         // We know this should be the right length (we just split it), and everything else would be
         // very unexpected.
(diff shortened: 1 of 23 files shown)


rosenpass/src/internal/cipher_traits/primitives/aead.rs
```diff
@@ -162,12 +162,12 @@ pub enum Error {
...
     #[error("internal error")]
-    InternalError,
+    Internal,
 
     /// Could not decrypt a message because the message is not a valid ciphertext for the given
     /// key.
-    #[error("decryption error")]
-    DecryptError,
+    #[error("decryption failed")]
+    DecryptionFailed,
```
SFeroudj-iteratec, line 170: I was also considering something like `InvalidCiphertext` here, based on the description in the doc comment. Any preferences?

PR #1082, opened 2026-09-09
Begin work on highly secure transports for the API

koraa:
This merges the work done in 2025 to introduce a security-focused communications backend for the shared memory API. The code has been rebased and cleaned up. It is preliminary work necessary before implementing the API itself.

In practice we are adding a lot generic of utilities for inter process communications, mostly focused around memfd-secret backed, memory-mapped buffers.

We can use the same utilities introduced here as a new back end for the secret memory allocation code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2010,6 +2010,7 @@ dependencies = [
  "assert_tv",
  "base64 0.23.1",
  "base64ct",
+ "bitflags 2.13.1",
  "blake2",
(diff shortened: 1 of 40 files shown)


ilka-schulz:
@koraa Should I review already? The PR is marked as ready (not a draft) and you requested a review. However, the CI fails and the PR is not scoped (labels, milestone, corresponding issue).

Issue #920, opened 2026-07-29
GH Action `regressions / multi-peer` builds binary again (probably not necessary)


ilka-schulz:

ilka-schulz:
- closed by #1085

PR #1085, opened 2026-09-14
ci: avoid duplicate rosenpass binary build for regressions workflow

etatlow-iteratec:
Issue #920

Introduces a shared binary build job for the regressions workflow jobs to avoid building the same binary twice


ilka-schulz (approved):
I went through all pitfalls I could think of and it looks good to me. In the future, we might want to build using `nix` and we might want to build for other workflows, too; but that is future work.

PR #1095, opened 2026-09-18
use badges.islabtech.com as a proxy for badges in the README

ilka-schulz:
Otherwise, we would leak the IP addresses of everyone reading the README to some badges providers, e.g. github.com, repology.org, shields.io.


koraa (approved):
Thank you! Can you please add an issue that we should move this to rosenpass infrastructure?

ilka-schulz:
> Thank you! Can you please add an issue that we should move this to rosenpass infrastructure?

I opened:
- #1100

